PR #23, opened 2015-01-10
Define the WebIDL2 property on self rather than window.

Ms2ger:
This makes it possible to run it in Workers.


darobin:
And released to npm, thanks!

PR #28, opened 2015-04-27
Extended Attribute rhs of `=` support string

othree:
Not in 1.0 spec or 2.0 draft.
But Mozilla is using this kind of syntax.

EX: https://github.com/mozilla/gecko-dev/blob/master/dom/webidl/BrowserElement.webidl

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -335,6 +335,9 @@
                 if (rhs = consume(ID)) {
                   ret.rhs = rhs
                 }
+                else if (rhs = consume(STR)) {
+                  ret.rhs = rhs
+                }
                 else if (consume(OTHER, "(")) {
                     rhs = [];
                     var id = consume(ID);


othree:
In the end, I found more issue of Mozilla Gecko's webidl.
There are [6 issues](https://github.com/othree/webidl2.js/commits/moz) including this one.
Only [one](https://github.com/othree/webidl2.js/commit/aee0292b0bf198052ce56586d3d1fa4d7071766e) of them is in WebIDL2 Draft.

I am not sure should I send all this changes back.
If you are ok to receive these non standard changes.
I will create another PR.

I can create a PR only contains the one in WebIDL2 Draft too.

darobin:
Sorry for not getting back to you earlier, I was travelling.

I would rather not include features that are Gecko-specific, especially since one of the use cases for this tool is to actually validate against the WebIDL specification. I would be open to discussing a special "Gecko-mode" that adds support for these, if it's helpful. I suspect however that that may be a moving target.

@bzbarsky Do you know if the differences between Gecko IDL and WebIDL are planned to be resolved?

othree:
No problem.
I will create another PR for the identifier in WebIDL 2 first.

bzbarsky:
@darobin In general, I'm not aware of such plans.  There are various Gecko-specific things that Gecko puts in its IDL that don't necessarily make sense to upstream since there would be no standard consumers.  This particular case (extended attribute with arbitrary string values) is one of those...

And yes, a gecko mode could be somewhat of a moving target, though we do generally try to not add syntax that's not already in the spec.

Looking at the list of six issues from https://github.com/othree/webidl2.js/commits/moz I think the status for the other five is:

1. jsonifier: this is just us avoiding what we consider the overcomplicated serializer syntax in Web IDL.  I suspect the spec will end up changing here at some point and then we will start tracking it.
2. Allowing '-' in identifiers:  This is part of Web IDL drafts already, not Mozilla-specific.
3. Trailing commas in enums: Part of Web IDL drafts already.  See https://www.w3.org/Bugs/Public/show_bug.cgi?id=17508
4. Attributes with sequence types: Will likely get added as https://www.w3.org/Bugs/Public/show_bug.cgi?id=23682 is resolved, but not standard IDL yet.
5. Empty interfaces (or more precisely what we call "external interfaces"):  Very much a Gecko-ism; it's used to declare non-WebIDL object types to the Web IDL system.  This will never end up in the spec, obviously.

PR #29, opened 2015-05-20
Add support for extended attributes identifier lists

tobie:
This adds support for constructs of the form:

```
[Exposed=(Window,Worker)]
```

which were kind of supported, but not documented.

The proposed JSON syntax for the above would be:

```json
{
    "name": "Exposed",
    "arguments": null,
    "rhs": {
        "type": "identifier-list",
        "value": [
            "Global",
            "Worker"
        ]
   }
}
```

This introduces a new Right Hand Side type, the [identifier list](http://heycam.github.io/webidl/#prod-ExtendedAttributeIdentList). I'm happy to change the JSON syntax, though, if so, please LMK.

Note that currently, the above would be rendered as:

```json
{
    "name": "Exposed",
    "arguments": null,
    "rhs": [
        "Global",
        "Worker"
    ]
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -660,9 +660,8 @@ The fields are as follows:
   whereas the lack thereof will yield a `null`. If there is an `rhs` field then
   they are the right-hand side's arguments, otherwise they apply to the extended
   attribute directly.
-* `rhs`: If there is a right-hand side, this will capture its `type` (always
-  "identifier" in practice, though it may be extended in the future) and its
-  `value`.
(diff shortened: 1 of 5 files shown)


jyasskin:
Looks fine to me.

tobie:
Oh, had missed your review here. Thanks!

Just want to make sure @darobin is confortable with the introduction of the `identifier-list` RHS type.

PR #31, opened 2015-05-23
In draft

othree:
Based on bzbarsky's [comment][1].
I create another pull request only contains changes already in draft.

* allow `-` in identifier
* allow trailing `,` in enum

[1]:https://github.com/darobin/webidl2.js/pull/28#issuecomment-100904710


lib/webidl2.js
```diff
@@ -6,7 +6,7 @@
         ,   re = {
                 "float":        /^-?(([0-9]+\.[0-9]*|[0-9]*\.[0-9]+)([Ee][-+]?[0-9]+)?|[0-9]+[Ee][-+]?[0-9]+)/
             ,   "integer":      /^-?(0([Xx][0-9A-Fa-f]+|[0-7]*)|[1-9][0-9]*)/
-            ,   "identifier":   /^[A-Z_a-z][0-9A-Z_a-z]*/
+            ,   "identifier":   /^[A-Z_a-z][0-9A-Z_a-z-]*/
```
jyasskin: @heycam, is this actually a good idea in [WebIDL](https://heycam.github.io/webidl/#idl-grammar)? It'll make identifiers that can't be used in Javascript. :-/

lib/webidl2.js
```diff
@@ -817,7 +817,7 @@
                 all_ws(store ? vals : null);
                 if (consume(OTHER, "}")) {
                     all_ws();
-                    if (saw_comma) error("Trailing comma in enum");
+                    // if (saw_comma) error("Trailing comma in enum");
```
jyasskin: When you're removing code, please just delete it. Don't comment it out.

darobin:
@heycam What was the reason for adding `-` to identifiers? As @jyasskin points out it seems like a pretty bad idea.

othree:
I think it is for CSS rules,ex `font-family`.

<https://github.com/othree/webidl/blob/master/webidl/CSS2Properties.webidl>

darobin:
@othree Ok, I still think it's a bad idea but since it's in the draft anyway... @othree Do you mind updating the code to be mergeable and to match @jyasskin's second comment please?

jyasskin:
This looks good to me.

jyasskin:
Although https://github.com/darobin/webidl2.js/pull/33/files got the trailing comma, so we can just skip that patch.

PR #45, opened 2016-06-22
Replace expect.js with expct

halton:
expect.js does not exist under node_modules/expect.js, where
test/web/make-web-test.js wrongly referenced. Besides, the expect.js
project (https://www.npmjs.com/package/expect.js) is not actively
maintained, expect (https://www.npmjs.com/package/expect) is an good
alternative library to be used.


marcoscaceres:
Merged this in manually

PR #55, opened 2017-01-09
Typecheck

markandrus:
Hi,

I find this library very useful, and I hope it will continue to be improved! The `README.md` was invaluable when consuming the AST, but I found myself wishing for type declarations, for example, to ensure I was handling every possible case. Also, as I began looking into `webidl2.js`'s source, I began to wonder if the code was indeed producing what the `README.md` claimed (without any comments, I found the code pretty difficult to read and I even spotted a few bugs).

So I made a few small cleanups, completed a first round of typechecking with Flow, then added Flow type declarations for the AST, then JSDoc, and finally TypeScript. (Please see the commit messages for information on these.) I hope these changes will ease consuming `webidl2.js` as well as development of `webidl2.js`. Here's an example of using the TypeScript declarations with YouCompleteMe:

screen shot 2017-01-08 at 11 45 22 pm

Would you consider such changes and additions?
Mark


markandrus:
I have pushed a few more changes in order to cut down on the duplication between the Flow, JSDoc, and TypeScript type declarations. I'm using (abusing?) WebIDL to declare webidl2.js's types, and then I use webidl2.js itself to parse the WebIDL and print out Flow, JSDoc, and TypeScript declarations. WebIDL is not expressive enough (AFAICT) to represent types like `[]`, `[Foo, Foo]`, and `"bar"`, so I am using dummy typedefs `EmptyArray` and `PairOfIDLTypes` to handle the former cases, and single-valued enums to handle the latter case. The code for this lives under `scripts/`.

marcoscaceres:
Hi @markandrus, sorry for the delay. I was waiting to get admin on this repo before responding (just got it!). I think this is a pretty cool idea... give me a few more days to look at it. Also, would you be willing to help me with maintenance? 

The code for this project was written pre-ES6, so it could use _a lot_ of love and modernization. Having said that, it does a really good job at parsing WebIDL. 

Anyway, if you are willing to commit to helping maintain this - and are willing to support any questions I might have wrt Flow (I've never used Flow - and am always a little scared of adding deps that could go away when the next hotness comes along) - then we can look at merging this.

markandrus:
Hi @marcoscaceres, thanks for taking a look!

> Also, would you be willing to help me with maintenance?

Assuming the maintenance is low-traffic, then yes I think I could commit, e.g., to reviewing PRs every now and again. I hesitate to commit to much more than that since my time is pretty stretched these days.

> The code for this project was written pre-ES6, so it could use a lot of love and modernization.

Yes, I was wondering, for example, could we break some of the functionality out into modules (would probably imply introducing a build step like Browserify for browsers), improving the tests, etc. Are there other tasks you've been thinking of?

>  and are willing to support any questions I might have wrt Flow (I've never used Flow - and am always a little scared of adding deps that could go away when the next hotness comes along) - then we can look at merging this.

FWIW, I'm not 100% sure Flow is the best choice going forward. I'm still relatively new to both TypeScript and Flow. I opted for Flow when introducing types here, since I knew they could be included inline in comments (or that entire sections could be skipped with, e.g., `//$FlowFixMe`). TypeScript would've been too radical a change in this PR... Although it might make more sense if we did add a build step. Have you used TypeScript?

Best,
Mark

marcoscaceres:
> Assuming the maintenance is low-traffic, then yes I think I could commit, e.g., to reviewing PRs every now and again. I hesitate to commit to much more than that since my time is pretty stretched these days.

We generally only update things once of twice a year. WebIDL is pretty stable and rarely gains new features.  

> Yes, I was wondering, for example, could we break some of the functionality out into modules (would probably imply introducing a build step like Browserify for browsers), improving the tests, etc. Are there other tasks you've been thinking of?

Exactly those:) And the coding style is very "early NodeJS" and super hard to decipher, lint, etc. This could do with some more declarative functional love.  

>FWIW, I'm not 100% sure Flow is the best choice going forward. I'm still relatively new to both TypeScript and Flow. I opted for Flow when introducing types here, since I knew they could be included inline in comments (or that entire sections could be skipped with, e.g., //$FlowFixMe). TypeScript would've been too radical a change in this PR... Although it might make more sense if we did add a build step. Have you used TypeScript?

So, I'm a bit worried about using things like TypeScript because generally those kinds of projects die after a while (been burned way too many times over last 20 years). I'd like to keep it vanilla - specially because this project only gets a few updates every few years and WebIDL is fairly stable. I'm hopeful that eventually TC-39 will come to its senses and add standardized types into JS... hopefully based on TypeScript, or Flow, the like, given that they've proven tremendously useful. 

Having said that, I really appreciate that you took the time to experiment adding types.

PR #61, opened 2017-04-03
Add support for annotated types

dontcallmedom:


marcoscaceres:
Will review tomorrow. Thanks for putting this together!

test/syntax/json/uniontype.json
```diff
@@ -1,87 +1,127 @@
...
-                    "union": true,
-                    "idlType": [
+   {
+      "partial" : false,
+      "members" : [
+         {
+            "idlType" : {
+               "idlType" : [
+                  {
+                     "array" : false,
```
marcoscaceres: What's outputting this that it's putting a space after the property name? Can we please not do that?

marcoscaceres: (eventually, we might want to generate this stuff using JSON.stringify() and we won't be able to match this format).

lib/webidl2.js
```diff
@@ -255,6 +255,13 @@
       return single_type() || union_type();
     };
 
+    var type_with_extended_attributes = function() {
```
marcoscaceres: Thankfully, this was not too painful 🏆

marcoscaceres:
Merge when ready 👍

PR #62, opened 2017-04-03
Remove support for MapClass (no longer valid in WebIDL)

dontcallmedom:


lib/webidl2.js
```diff
@@ -370,16 +358,12 @@
...
-        else if (pair = type_pair()) {
-          ret.typePair = pair;
-        }
         // [Constructor()]
         else {
           ret["arguments"] = [];
         }
         all_ws();
-        consume(OTHER, ")") || error("Unexpected token in extended attribute argument list or type pair");
+        consume(OTHER, ")") || error("Unexpected token in extended attribute argument list");
```
marcoscaceres: Can you include the erroneous token in the error? `": " + token)`. Makes it easier when reporting issues.

marcoscaceres: Ok, thanks for clarifying

PR #67, opened 2017-05-15
Remove deprecated IDL arrays and exceptions

### LINKED ISSUE #65: Remove exceptions support
webidl2.js includes [a function for parsing exceptions](https://github.com/w3c/webidl2.js/blob/a0bccdf485a511a94042d8d541a9cfbcc7de15c3/lib/webidl2.js#L919) (I even went so far as to type these as `IDLException` in [my typed branch](https://github.com/w3c/webidl2.js/pull/55)); however, upon closer inspection, I don't believe WebIDL allows exceptions to be declared. See [here](https://www.w3.org/TR/2016/PR-WebIDL-1-20160915/#idl-exceptions): 

> An **exception** is a type of object that represents an error and which can be thrown or treated as a first class value by implementations. Web IDL does not allow exceptions to be defined, but instead has a number of pre-defined exceptions that specifications can reference and throw in their definition of operations, attributes, and so on.

I also don't see any reference to "exception" in the grammar, so I think this code should be removed (originally introduced [here](https://github.com/w3c/webidl2.js/commit/da8ef5ea208a981b99d60bcf54c64b408175490a)).

### LINKED ISSUE #66: Is array syntax dead?
The `type_suffix` function parses array syntax (`type[]`) but I cannot find a relevant reference on Web IDL spec, is it dead?

```
Type ::
    SingleType
    UnionType Null
```

SaschaNaz:
Fixes #65
Fixes #66


SaschaNaz:
[`serializer` is also at risk](https://github.com/heycam/webidl/pull/323), probably can be covered later.

marcoscaceres:
I'll release this later today with a mayor version bump.

marcoscaceres:
Released as 3.0

PR #69, opened 2017-05-27
Test for latest LTS/stable node versions

SaschaNaz:
1. Add Node LTS and stable branches to CI
2. Remove unused dependencies (specifically `microtime` fails to build on Node 7 with unknown reason)


marcoscaceres:
thanks again, @SaschaNaz!

PR #75, opened 2017-07-01
Support TypeWithExtendedAttributes on generics

SaschaNaz:
... which includes `sequence`, `FrozenArray` and `record`. `Promise` is intentionally not supported because of [the spec](https://heycam.github.io/webidl/#idl-types):

```
PromiseType ::
    Promise < ReturnType >
```

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -195,7 +195,7 @@
           var types = [];
           do {
             all_ws();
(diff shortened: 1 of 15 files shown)


marcoscaceres:
published as 4.1.0. Thanks again, @SaschaNaz! You are a superstar!

PR #76, opened 2017-07-17
Add "sequence" property to IDL Type AST definition

lerouche:
When IDL Type syntax is parsed, its resulting AST object has a property called "sequence". This is declared in the README everywhere except the IDL Type section.


marcoscaceres:
Thanks!

PR #80, opened 2017-09-21
Check duplicated names

### LINKED ISSUE #77: Should report error for using duplicate names
Consider the following example idl:

```webidl
typedef int Test;

interface Test {
  void foo();
};

```

The generated output is as
```json
[
    {
        "extAttrs": [],
        "idlType": {
            "generic": null,
            "idlType": "int",
            "nullable": false,
            "sequence": false,
            "union": false
        },
        "name": "Test",
        "type": "typedef"
    },
    {
        "extAttrs": [],
        "inheritance": null,
        "members": [
            {
                "arguments": [],
                "creator": false,
                "deleter": false,
                "extAttrs": [],
                "getter": false,
                "idlType": {
                    "generic": null,
                    "idlType": "void",
                    "nullable": false,
                    "sequence": false,
                    "union": false
                },
                "legacycaller": false,
                "name": "foo",
                "setter": false,
                "static": false,
                "stringifier": false,
                "type": "operation"
            }
        ],
        "name": "Test",
        "partial": false,
        "type": "interface"
    }
]
```

In this case, the name `Test` is used twice: once for the interface and once for the typedef.

Should the library report this as an error?

SaschaNaz:
Fixes #77.

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -42,7 +42,7 @@
   var parse = function(tokens, opt) {
     var line = 1;
     tokens = tokens.slice();
-    var names = new Set();
+    var names = new Map();
 
     var FLOAT = "float",
       INT = "integer",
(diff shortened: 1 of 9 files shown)


SaschaNaz:
I have a question, do we have to keep ES5 compatibility or can we use ES2015 now?

marcoscaceres:
It’s 2107, I say we embrace the future... and traspile our selves out of trouble if need be.

SaschaNaz:
That will be good, I'll try it.

marcoscaceres:
@SaschaNaz, can you please add yourself to the contributors at https://github.com/w3c/webidl2.js/blob/develop/package.json

PR #84, opened 2017-10-16
BREAKING CHANGE: Use ES2015 syntax

### LINKED ISSUE #82: Drop support for Node 4, move to 6 LTS
In October, we should drop Node 4, and start targeting. 

It might also be a good chance to rewrite this is ES6. 

Would like to be part of that effort so I can learn the code base a bit better.

SaschaNaz:
This is related to #82 but not really breaking compat with Node.js 4.

This PR converts:

* `var func = function() { }` style to `function func() { }`
* `var x` to `let x` or `const x` depending on the variable is reassigned or not
* ES5-style class declaration for WebIDLParseError to ES2015 one
* `for (var i = 0; i < array.length; i++)` to `for (const item of array)`
* callback functions to use arrow syntax
* object literal member functions `{ func: function () { } }` to `{ func() { } }`

PS: Now fixes #82 by removing it on `.travis.yml`.

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -1,21 +1,22 @@
+"use strict";
+
 (function() {
-  var tokenise = function(str) {
-    var tokens = [],
-      re = {
-        "float": /^-?(([0-9]+\.[0-9]*|[0-9]*\.[0-9]+)([Ee][-+]?[0-9]+)?|[0-9]+[Ee][-+]?[0-9]+)/,
-        "integer": /^-?(0([Xx][0-9A-Fa-f]+|[0-7]*)|[1-9][0-9]*)/,
-        "identifier": /^[A-Z_a-z][0-9A-Z_a-z-]*/,
-        "string": /^"[^"]*"/,
-        "whitespace": /^(?:[\t\n\r ]+|[\t\n\r ]*((\/\/.*|\/\*(.|\n|\r)*?\*\/)[\t\n\r ]*))+/,
-        "other": /^[^\t\n\r 0-9A-Z_a-z]/
-      },
-      types = ["float", "integer", "identifier", "string", "whitespace", "other"];
+  function tokenise(str) {
+    const tokens = [];
+    const re = {
+      "float": /^-?(([0-9]+\.[0-9]*|[0-9]*\.[0-9]+)([Ee][-+]?[0-9]+)?|[0-9]+[Ee][-+]?[0-9]+)/,
+      "integer": /^-?(0([Xx][0-9A-Fa-f]+|[0-7]*)|[1-9][0-9]*)/,
+      "identifier": /^[A-Z_a-z][0-9A-Z_a-z-]*/,
+      "string": /^"[^"]*"/,
+      "whitespace": /^(?:[\t\n\r ]+|[\t\n\r ]*((\/\/.*|\/\*(.|\n|\r)*?\*\/)[\t\n\r ]*))+/,
+      "other": /^[^\t\n\r 0-9A-Z_a-z]/
+    };
+    const types = ["float", "integer", "identifier", "string", "whitespace", "other"];
     while (str.length > 0) {
-      var matched = false;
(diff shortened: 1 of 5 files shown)


lib/webidl2.js
```diff
@@ -27,51 +28,53 @@
...
+      this.input = input;
+      this.tokens = tokens;
+    }
 
-  WebIDLParseError.prototype.toString = function() {
-    return this.message + ", line " + this.line + " (tokens: '" + this.input + "')\n" +
-      JSON.stringify(this.tokens, null, 4);
-  };
+    toString() {
+      return this.message + ", line " + this.line + " (tokens: '" + this.input + "')\n" +
```
marcoscaceres: Maybe use template literal here?

lib/webidl2.js
```diff
@@ -27,51 +28,53 @@
...
-      var maxTokens = 5;
+    function error(str) {
+      let tok = "";
+      let numTokens = 0;
+      const maxTokens = 5;
       while (numTokens < maxTokens && tokens.length > numTokens) {
         tok += tokens[numTokens].value;
         numTokens++;
       }
       throw new WebIDLParseError(str, line, tok, tokens.slice(0, 5));
```
marcoscaceres: Maybe we should fix this now:

```
throw new WebIDLParseError(str, line, tok, tokens.slice(0, maxTokens))
```

lib/webidl2.js
```diff
@@ -714,15 +718,15 @@
           consume(OTHER, ";") || error("Missing semicolon after interface");
           return ret;
         }
-        var ea = extended_attrs(store ? mems : null);
+        const ea = extended_attrs(store ? mems : null);
```
marcoscaceres: nit: whitespace ?

marcoscaceres: Be nice if we got rid of the `const x = {}, y = {}; ` etc. pattern usage and instead do "const x = {}; const y = {}" and so on. The "," is a pretty bad practice. We can do that in different PR tho.

marcoscaceres:
We need to update the travis file as per #82  (i.e., drop node 4):

```
node_js:
  - "lts/*"
  - "6"
```

I don't know if this works, but we could try:

```
node_js:
  - "lts/*"
  - "lts/6"
```

lib/webidl2.js
SaschaNaz: >nit: whitespace ?

Could you clarify more? `extended_attrs` already calls `all_ws` internally so no need to process whitespaces twice.

marcoscaceres: sorry, I meant changing:

```
const mems = [],
  ret = {};
```

To:
```
const mems = [];
const ret = {};
```

marcoscaceres: With regards to whitespace.... I literally meant that the indentation of your code is incorrect - just a nit :) The actual processing is  all good 👍

SaschaNaz:
>We need to update the travis file as per #82 (i.e., drop node 4):

Not adding `node` will exclude non-LTS latest stable versions (9, 11, ...), do we want to exclude them?

marcoscaceres:
> Not adding node will exclude non-LTS latest stable versions (9, 11, ...), do we want to exclude them?

Ah, good point. Yeah, we probably want to test whatever is latest. I have to run off for a bit, can you please check travisci docs for how to do that?

SaschaNaz:
https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Specifying-Node.js-versions

>`node` latest stable Node.js release

Current `node` points to the latest version. Not sure whether `lts/6` will work or not but it's probably not good to use an undocumented string.

SaschaNaz:
`lts/6` didn't work but [the documentation says `lts/boron` work](https://github.com/creationix/nvm#long-term-support).

marcoscaceres:
Please squash merge and use "BREAKING CHANGE: Use ES2015 syntax" as commit message.

PR #88, opened 2017-10-17
Use ES2015 syntax for tests

SaschaNaz:
Looking at the tests I thought they should be rewritten in ES2015 before any possible fix here.

Additionally this removes lib-cov references that don't exist.

### DIFF
--- a/test/invalid.js
+++ b/test/invalid.js
@@ -1,42 +1,40 @@
-
 // NOTES:
 //  - the errors actually still need to be reviewed to check that they
 //    are fully correct interpretations of the IDLs
 
-var wp = process.env.JSCOV ? require("../lib-cov/webidl2") : require("../lib/webidl2")
-,   expect = require("expect")
-,   pth = require("path")
-,   fs = require("fs")
-;
-describe("Parses all of the invalid IDLs to check that they blow up correctly", function () {
-    var dir = pth.join(__dirname, "invalid/idl")
-    ,   skip = {}
-    ,   idls = fs.readdirSync(dir)
-                  .filter(function (it) { return (/\.w?idl$/).test(it) && !skip[it]; })
-                  .map(function (it) { return pth.join(dir, it); })
-    ,   errors = idls.map(function (it) { return pth.join(__dirname, "invalid", "json", pth.basename(it).replace(/\.w?idl/, ".json")); })
-    ;
+"use strict";
 
-    for (var i = 0, n = idls.length; i < n; i++) {
-        var idl = idls[i], error = JSON.parse(fs.readFileSync(errors[i], "utf8"));
-        var func = (function (idl, err) {
-            return function () {
-                var error;
-                try {
-                    var ast = wp.parse(fs.readFileSync(idl, "utf8"));
-                    console.log(JSON.stringify(ast, null, 4));
-                }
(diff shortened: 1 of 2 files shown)


test/invalid.js
```diff
@@ -1,42 +1,40 @@
...
+"use strict";
+
+const wp = require("../lib/webidl2");
+const expect = require("expect");
+const pth = require("path");
+const fs = require("fs");
+
+describe("Parses all of the invalid IDLs to check that they blow up correctly", () => {
+  const dir = pth.join(__dirname, "invalid/idl");
+  const skip = {};
```
marcoscaceres: the `skip` should probably be a `new Set()`. I'm not sure `skip` very useful tho, as all the names are going to be unique, right?

test/syntax.js
```diff
@@ -1,41 +1,39 @@
...
+const pth = require("path");
+const fs = require("fs");
+const jdp = require("jsondiffpatch");
+const debug = true;
 
-    for (var i = 0, n = idls.length; i < n; i++) {
-        var idl = idls[i], json = jsons[i];
+describe("Parses all of the IDLs to produce the correct ASTs", () => {
+  const dir = pth.join(__dirname, "syntax/idl");
+  const skip = {}; // use if we have a broken test
```
marcoscaceres: as above, about `skip`.

test/invalid.js
```diff
@@ -1,42 +1,40 @@
...
-                }
-                catch (e) {
-                    error = e;
-                }
-                finally {
-                    expect(error).toBeTruthy();
-                    expect(error.message).toEqual(err.message);
-                    expect(error.line).toEqual(err.line);
-                }
+  for (let i = 0, n = idls.length; i < n; i++) {
```
marcoscaceres: Nit: 
```
for(const idl of idl) {

}
```

marcoscaceres: fixed typo in code above

SaschaNaz: Couldn't do it because of `errors[i]`.

marcoscaceres: ah, ok

test/invalid.js
SaschaNaz: Probably an array will be sufficient enough as `new Set()` will eventually be written as  `new Set(["testname1", "testname2"])`.

>I'm not sure skip very useful tho, as all the names are going to be unique, right?

I'm sorry but not sure why the uniqueness makes `skip` not useful, can you clarify more?

marcoscaceres: I guess the question is, under what circumstances will there be duplicate file names? (Or have I misunderstood what `skip` doing?) If the answer is "never", then the uniqueness check is not needed.

SaschaNaz: Well, I think `skip` is for temporarily disabling a test which cannot be fixed in the short term, or just a test-to-do-later.

marcoscaceres: I'm ok to keep it as is. If you are happy with it, I'm happy with it :)

SaschaNaz: I think the uniqueness is even an assumption for `skip` as it would accidentally skip multiple tests if names are not unique.

marcoscaceres:
published as webidl2@6.0.1

PR #91, opened 2017-10-23
WIP: error if dict has NoInterfaceObject

### LINKED ISSUE #81: [NoInterfaceObject] on Dictionary types
We should probably abort if we see `[NoInterfaceObject]` on a Dictionary. Just saw this in spec I was reviewing.

marcoscaceres:
* closes #81

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -13,6 +13,9 @@
       "whitespace": /^(?:[\t\n\r ]+|[\t\n\r ]*((\/\/.*|\/\*(.|\n|\r)*?\*\/)[\t\n\r ]*))+/,
       "other": /^[^\t\n\r 0-9A-Z_a-z]/
     };
+    const extAttrsNotAllowedOnDict = new Set([
+      "NoInterfaceObject"
+    ])
     const types = ["float", "integer", "identifier", "string", "whitespace", "other"];
     while (str.length > 0) {
       let matched = false;
@@ -823,6 +826,13 @@
       all_ws(isPartial ? null : store, "pea");
       if (!consume(ID, "dictionary")) return;
       all_ws();
+      const notAllowed = Array.isArray(store) && store
+        .filter(({ name }) => extAttrsNotAllowedOnDict.has(name))
+        .map(({ name }) => name)
+        .join(", ");
+      if (notAllowed) {
+        error(`Found extended attribute(s) not allowed on dictionary: ${notAllowed}.`);
+      }
       const name = consume(ID) || error("No name for dictionary");
       const mems = [];
       const ret = {
--- a/test/invalid/idl/dict-nointerfaceobject.idl
+++ b/test/invalid/idl/dict-nointerfaceobject.idl
@@ -1,0 +1,3 @@
+// NoInterfaceObject can't appear on a Dictionary
+[NoInterfaceObject]
+dictionary Dict {};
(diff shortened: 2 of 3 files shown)


marcoscaceres:
this is not working yet... @TimothyGu or @SaschaNaz, could either of you take a look? It's saying that `received` is undefined in the test:

```
Error: expect(received).toBeTruthy()
```

Any ideas on what I'm doing wrong?

lib/webidl2.js
```diff
@@ -13,6 +13,9 @@
       "whitespace": /^(?:[\t\n\r ]+|[\t\n\r ]*((\/\/.*|\/\*(.|\n|\r)*?\*\/)[\t\n\r ]*))+/,
       "other": /^[^\t\n\r 0-9A-Z_a-z]/
     };
+    const extAttrsNotAllowedOnDict = new Set([
```
SaschaNaz: Scope problem here, this should be defined in `parse()`.

lib/webidl2.js
```diff
@@ -823,6 +826,13 @@
       all_ws(isPartial ? null : store, "pea");
       if (!consume(ID, "dictionary")) return;
       all_ws();
+      const notAllowed = Array.isArray(store) && store
```
SaschaNaz: `store` is only defined with `ws` option (`parse(text, { ws: true })`), and its purpose is to collect whitespaces. It's totally not being tested and is undocumented.

marcoscaceres: heh, whoops :) Ok, but it's the thing that contains the extended attribute objects? Would we need to move those somehow?

SaschaNaz: No, it contains previous top level definitions including interfaces, enums, etc., so that any whitespaces between them can be mixed in the list.

TimothyGu:
I'm not sure this is a good idea… in the spec NoInterfaceObject is only defined for the ECMAScript language binding, while I would assume this parser to be binding-agnostic. I understand that currently JS is the only language binding spec'd (not counting Java) but I remember there were plans to add WebAssembly support to Web IDL.

marcoscaceres:
@TimothyGu consider that `NoInterfaceObject` is is exclusively for interfaces (irrespective of binding language). It's also an "undesirable feature" as per the spec, so I doubt we would see it used in any other language binding.

lib/webidl2.js
SaschaNaz: I think we should explicitly pass extattrs list to `dictionary()`.

https://github.com/w3c/webidl2.js/blob/96943e30d486ed949cbfe0e8315da92f9b114910/lib/webidl2.js#L967

This way `definition()` should receive it first and pass it to dictionary parser.

marcoscaceres: Yeah, that would be nice actually. I noticed it called the function twice.

SaschaNaz: BTW, shouldn't it be *error if anything else but an interface has NoInterfaceObject*?

marcoscaceres: Yeah, I was thinking the same. Might need to rewrite all this to clarify things for each type ("interfaces shouldn't have of these: x, y, z", "dictionaries shouldn't have any of these: w,z", and so on).

SaschaNaz:
We probably can make this optional with something like `.parse(text, { binding: "javascript" })`.

marcoscaceres:
I’m working on an independent IDL validator in ReSpec, but I’ll make it generic so we can add it to this project. I’ll spit out a list of warnings as markdown.

Issue #94, opened 2017-10-26
Type conversion on default values is destructive 


marcoscaceres:
Given 

```
interface X {
  const unsigned long BIT_MASK = 0x0000fc00;
}
```

Becomes:

```
"{
  "type": "number",
  "value": 64512
}"
```

I think the only fix here is not to do the number conversion. 

@SaschaNaz, thoughts?

SaschaNaz:
👍 

```json
{
  "type": "number",
  "value": 64512,
  "raw": "0x0000fc00"
}
```

Alternatively we can add `base: 16` thing but that won't recover preceding zeros, which may be a problem in some situation.

marcoscaceres:
I like the “raw” member, but I’m still not entirely sure of the use case for doing the conversion during parse. It seems that if a consumer has type “number” and “value”, they can perform the conversion themselves.

marcoscaceres:
Apologies, I see you answered the above in the PR. 🙏

SaschaNaz:
Yeah, the conversion is reasonably easy so it shouldn't be a big issue for users who uses the resulting number value for whatever reason.

PR #95, opened 2017-10-26
 BREAKING CHANGE: argument + default types should be string

marcoscaceres:


marcoscaceres:
It's weird that the this had not been caught earlier.

gsnedders:
What was the reason for this change? How should I get the value as a JS Number primitive? Is there any danger in doing `1 * ret.value` as was done before?

Perhaps most vitally, given a value that isn't representable as a JS Number, does the rounding happen in the right way to match what would happen with a `const` in a browser?

saschanaz:
> What was the reason for this change?

This is because of #94, where the parsed result does not provide the original number literals. `1 * ret.value` should give the exact same value as the one previously provided.

PR #102, opened 2017-11-02
chore: drop Node 6 support

marcoscaceres:
Follow latest stable and LTS going forward.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,4 +2,3 @@ language: node_js
 node_js:
   - node
   - lts/*
-  - 6


TimothyGu:
@marcoscaceres Would you please reconsider this decision? According to the [release schedule](https://github.com/nodejs/Release#release-schedule) of Node.js, v6.x will remain in LTS through April of 2018. Since v6.x already has full support for ES2015, I don't expect too much (if any) additional work to come out of maintaining compatibility with v6.x.

I am a de-facto maintainer for [webidl2js](https://www.npmjs.com/package/webidl2js), which follows [jsdom](https://www.npmjs.com/package/jsdom)'s pace on Node.js version support -- which is currently all active LTS branches + Current branch. Considering webidl2js composes 40% of the usage of webidl2.js by monthly downloads, it would be nice for this to be reverted.

marcoscaceres:
I'm ok with that. We don't use any async/await and only a few limited fancy things, so it should be fine. However, can you check and let us know if we need to revert any recent code changes (will make you in charge of keeping Node6 compat :)). 

Also, @TimothyGu, I'll make you an official maintainer. Your participation has been invaluable thus far. I'll send you an invite.

marcoscaceres:
Actually, it seems you already part of the team 👍 Please add your details to the package.json file.

TimothyGu:
@marcoscaceres 

> However, can you check and let us know if we need to revert any recent code changes (will make you in charge of keeping Node6 compat :)).
>
> Please add your details to the package.json file.

I'd be happy to do both of those :)

PR #105, opened 2017-11-02
Support mixins + includes statements

### LINKED ISSUE #92: Add mixin support
Seems we don't support [mixins](https://heycam.github.io/webidl/#idl-interface-mixins) yet, right?

```
interface mixin Observable {
  void addEventListener(DOMString type,
                        EventListener listener,
                        boolean useCapture);
  // ...
};
```

SaschaNaz:
(This includes #103)

Fixes #92.

[MixinMember](https://heycam.github.io/webidl/#prod-MixinMember) only includes stringifiers but not static members, while current `operation()`/`attribute()` always include both. This PR pulls `static_member()` and `stringifier()` out of `operation()`/`attribute()` to support mixins + better match the parser structure to the spec.

```
MixinMember ::
    Const
    RegularOperation
    Stringifier
    ReadOnly AttributeRest
```

Bonus: This renames nonspecial_operation to regular_operation to match the spec naming.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,7 @@ Where the fields are as follows:
 * `union`: Boolean indicating whether this is a union type or not.
 
 ### Interface
+
 Interfaces look like this:
 
 ```JS
@@ -167,7 +168,36 @@ The fields are as follows:
   sense.
 * `extAttrs`: A list of [extended attributes](#extended-attributes).
 
+### Interface mixins
+
+Interfaces mixins look like this:
+
+```JS
+{
+  "type": "interface-mixin",
+  "name": "Animal",
+  "partial": false,
+  "members": [...],
+  "extAttrs": [...]
+}, {
+  "type": "interface-mixin",
+  "name": "Human",
+  "partial": false,
+  "members": [...],
+  "extAttrs": [...]
+}
+```
+
+The fields are as follows:
+
+* `type`: Always "interface-mixin".
+* `name`: The name of the interface mixin.
+* `partial`: A boolean indicating whether it's a partial interface mixin.
+* `members`: An array of interface members (attributes, operations, etc.). Empty if there are none.
+* `extAttrs`: A list of [extended attributes](#extended-attributes).
+
 ### Namespace
+
 Namespaces look like this:
 
 ```JS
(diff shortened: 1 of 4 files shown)


lib/webidl2.js
```diff
@@ -569,14 +568,7 @@
...
       all_ws(store, "pea");
-      const ret = {
-        type: "operation",
-        getter: false,
-        setter: false,
-        deleter: false,
-        "static": false,
-        stringifier: false
-      };
+      const ret = Object.assign({}, EMPTY_OPERATION);
```
marcoscaceres: this is nice :)

lib/webidl2.js
```diff
@@ -624,6 +603,29 @@
...
+      all_ws();
+      return noninherited_attribute(store, prefix) ||
+        regular_operation(store, prefix) ||
+        error("No body in static member");
+    }
+
+    function stringifier(store) {
+      all_ws(store, "pea");
+      if (!consume(ID, "stringifier")) return;
+      const prefix = "stringifier";
```
marcoscaceres: Maybe just inline this?

lib/webidl2.js
```diff
@@ -863,7 +910,7 @@
       }
     };
 
-    const enum_ = function(store) {
+    const enum_ = function (store) {
```
marcoscaceres: nit: whitespace after `function`

lib/webidl2.js
SaschaNaz: Done!

lib/webidl2.js
SaschaNaz: Oops, the VSCode formatter 😬

marcoscaceres: We should add `.vstudio` file at the base... I'm also using code, and we should start formatting this stuff properly.

SaschaNaz: Wait, it should be `function enum_` instead of `const enum_`. Omitted on the previous PR.

marcoscaceres:
@SaschaNaz, if you are happy to merge this, I'm happy to ship it and add support in ReSpec :) Let me know.

README.md
```diff
@@ -167,7 +168,36 @@ The fields are as follows:
   sense.
 * `extAttrs`: A list of [extended attributes](#extended-attributes).
 
+### Interface mixins
+
+Interfaces mixins look like this:
+
+```JS
+{
+  "type": "interface-mixin",
```
marcoscaceres: I'm wondering, maybe we should change this to "interface mixin"

marcoscaceres: (that would match "callback interface", and would allow me to use the type directly in ReSpec without needing to special case the type.

SaschaNaz: This is to keep consistency with `line-comment`, `multiline-comment`, etc., although they are undocumented.

SaschaNaz: (Do we want to change them as `line comment`, `multiline comment`, etc.?)

marcoscaceres:
Happy to also make this change, let me know @SaschaNaz.

README.md
marcoscaceres: Nah, line-comment and multiline-comment are ok.

SaschaNaz: Change made! (But why not `line comment`?)

SaschaNaz:
I don't see any problems now so it should be safe to merge 😁

README.md
marcoscaceres: Only because that would be a breaking change (e.g., in ReSpec, I have code that already expects "line-comment").

marcoscaceres:
+ webidl2@8.0.0

PR #110, opened 2017-11-03
Always add rhs property

### LINKED ISSUE #96: Extended Attributes `rhs` should always be there
It's somewhat annoying to have to check for missing properties, as is the case for `rhs` on extended attributes. 

We should make sure we don't conditionally add/remove members - and where they are null, they should just be `null`.

SaschaNaz:
Fixes #96.

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -343,32 +343,28 @@
         name: name.value,
         "arguments": null,
         type: "extended-attribute",
+        rhs: null
       };
       all_ws();
       const eq = consume(OTHER, "=");
       if (eq) {
-        let rhs;
         all_ws();
-        if (rhs = consume(ID)) {
-          ret.rhs = rhs;
-        } else if (rhs = consume(FLOAT)) {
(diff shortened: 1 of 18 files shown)


lib/webidl2.js
```diff
@@ -343,32 +343,28 @@
         name: name.value,
         "arguments": null,
         type: "extended-attribute",
+        rhs: null
```
marcoscaceres: I'm now wondering if we should just make it an empty list?

marcoscaceres: actually, ignore this.

marcoscaceres:
Shipped at 8.1.0 :)

SaschaNaz:
>Generally speaking, and from a functional perspective, it's quite frustrating to deal with when you sometimes get nothing, sometimes an object, other times an array :(

I agree, could you file an issue for that?

marcoscaceres:
filed: 
https://github.com/w3c/webidl2.js/issues/113

PR #115, opened 2017-11-13
 Revert "chore: drop Node 6 support (#102)"

TimothyGu:
Per https://github.com/w3c/webidl2.js/pull/102#issuecomment-343852600. Also some metadata updates.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,3 +2,4 @@ language: node_js
 node_js:
   - node
   - lts/*
+  - 6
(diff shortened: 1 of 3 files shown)


marcoscaceres:
@TimothyGu, should we hold doing a release until we do some additional bug fixes? This changes here are mostly just for us and don't affect folks downstream.

TimothyGu:
@marcoscaceres I'm fine with leaving it unreleased.

PR #129, opened 2018-01-19
Let error messages include the current definition name

### LINKED ISSUE #119: Improve "No semicolon after enum" message
Would be great if we could improve "No semicolon after enum" to "Missing semicolon after ${identifier} enum".

saschanaz:
Fixes #119 

This should be simple enough 😁

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -495,7 +495,7 @@
       all_ws();
       const tok = consume(ID, "interface");
       if (tok) {
-        ret = interface_rest();
+        ret = interface_rest(false);
         ret.type = "callback interface";
         return ret;
       }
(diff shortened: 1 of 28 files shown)


marcoscaceres:
Should we deal with partials?

marcoscaceres:
That is, should the error say, for example, "`partial interface NoSemicolon`: Missing semicolon after interface" when given a partial?

saschanaz:
Oh, partial definitions... It should also be simple enough, so yeah I will push one more commit.

marcoscaceres:
Ignore my comment above. I see it's handled by `.type`. Maybe add a test for partial?

marcoscaceres:
Oh, wait, maybe it's not handled by type... it must be Friday!

saschanaz:
Callback interfaces should also be supported, currently an error will call it just an interface.

saschanaz:
Friday commits added!

marcoscaceres:
+ webidl2@10.1.0

Issue #93, opened 2017-10-25
types of types


marcoscaceres:
For interfaces, const, attributes etc. that define a idlType, those types should themselves have a `type:` property saying what they are (e.g., "attribute-type", "constant-type").

SaschaNaz:
I'm starting a bit confused about the hyphening, when we want to use space and when we want to use hyphen for the type name?

SaschaNaz:
And what should the name be for returns types? Just `callback-type` or `callback-return-type`?

SaschaNaz:
And BTW, `iterable()` has an idlType or an array of idlTypes. Do we want to make it a consistent array?

marcoscaceres:
Ok, so the rule appears to be: if the type can match a name for display purpose (e.g., “callback interface”, “interface mixin”) we use space, otherwise use hyphen. 

And yeah about iterable, also setlike should be an array with a single item.

saschanaz:
Sometimes an `idlType` has an `idlType` that includes an `idlType` with an `idlType`:

```json
"arguments": [
    {
        "optional": false,
        "variadic": false,
        "extAttrs": [],
        "idlType": {
            "sequence": true,
            "generic": "sequence",
            "nullable": false,
            "union": false,
            "idlType": {
                "sequence": false,
                "generic": "record",
                "nullable": false,
                "union": false,
                "idlType": [
                    {
                        "sequence": false,
                        "generic": null,
                        "nullable": false,
                        "union": false,
                        "idlType": "ByteString"
                    },
                    {
                        "sequence": false,
                        "generic": null,
                        "nullable": false,
                        "union": false,
                        "idlType": "any"
                    }
                ]
            }
        },
        "name": "param"
    }
]
```

Do we want to also add `sequence-type`, `record-type`, etc. for them? How about `union-type`?

marcoscaceres:
that would be great, actually. That's part what I'm doing at the moment with my dynamically adding a parent (in ReSpec) to work out the type's type.

saschanaz:
That sounds essentially #85 😄. Can you point your related ReSpec commits if you pushed so that I can understand your use case better?

marcoscaceres:
I've not pushed my local the code up yet, but essentially I am deriving the "type" for a data-attribute. For example:

```HTML
<a dataset-idl-type="attribute-type">DOMString</a>
```

So, to do that, I call `getTypeFromParent()`, which just combines into, for example, "attribute-type": 

```JS
[...]

PR #130, opened 2018-01-26
Auto acquisition for parser result changes

saschanaz:
I tried fixing #93 and got a BIG ~~AST~~ parser result changes, so I created a tool to automate change acquisitions.

Much shorter test scripts as a bonus!

Also I fixed an issue that `typedef-union` test was being ignored because of its file extension.


test/invalid.js
```diff
@@ -4,38 +4,17 @@
...
-      }
-      catch (e) {
-        error = e;
-      }
-      finally {
-        expect(error).toBeTruthy();
-        expect(error.message).toEqual(err.message);
-        expect(error.line).toEqual(err.line);
-      }
+  for (const test of collect("invalid", true)) {
```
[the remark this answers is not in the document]
marcoscaceres: Actually, let's fix this at the source.

test/util/collect.js
```diff
@@ -0,0 +1,61 @@
...
+const pth = require("path");
+const fs = require("fs");
+const jdp = require("jsondiffpatch");
+
+/**
+ * Collects test items from the specified directory
+ * @param {string} base
+ * @param {boolean} [expectError]
+ */
+function* collect(base, expectError) {
```
marcoscaceres: Let's make `expectError` not be a boolean. Maybe make it an object, so to avoid the boolean trap?

test/util/collect.js
```diff
@@ -0,0 +1,61 @@
...
+  }
+};
+
+/**
+ * Creates a test item object
+ * @param {string} ast
+ * @param {string} path
+ * @param {Error} [error]
+ */
+function createItem(ast, path, error) {
```
marcoscaceres: Could be a `class TestItem` maybe? Might make this a little cleaner.

test/util/collect.js
```diff
@@ -0,0 +1,61 @@
...
+  return {
+    ast,
+    path,
+    error,
+    jsonPath() {
```
marcoscaceres: nit: maybe `get jsonPath()`? then just `item.jsonPath;`

test/util/collect.js
```diff
@@ -0,0 +1,61 @@
...
+      return pth.join(pth.dirname(this.path), "../json", pth.basename(this.path).replace(".widl", ".json"));
+    },
+    diff() {
```
marcoscaceres: Nit: I guess if we are only using this with testing it doesn't matter too much about using `readFileSync`... or could we spin up the tests to run in parallel?

saschanaz:
Great reviews! I applied your feedbacks, although I'm not sure I understand this comment:

>Nit: I guess if we are only using this with testing it doesn't matter too much about using readFileSync... or could we spin up the tests to run in parallel?

The `readFileSync` part is now moved to `readJSON()`, but should we use async `fs.readFile`? Things will be easy enough if we use Node 8+ as it supports async functions.

marcoscaceres:
> The readFileSync part is now moved to readJSON(), but should we use async fs.readFile? Things will be easy enough if we use Node 8+ as it supports async functions.

I think even Node 6 supported promises, but obviously not `await` (so, not as pretty)... having said that, let's leave it reading the files synchronously. The tests run pretty quick already :)

PR #131, opened 2018-01-29
feat: add types on idlTypes

### LINKED ISSUE #93: types of types

saschanaz:
Fixes #93 

(Happy one-tap acquisition :tada:)

I added only `attribute-type` to get some feedback first. Consts (or anything that call `primitive_type()`) define idlType as mere strings, do we want to make them objects with `type: "const-type"`?

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -208,9 +208,9 @@
       }
     };
 
-    function single_type() {
+    function single_type(typeName) {
       const prim = primitive_type();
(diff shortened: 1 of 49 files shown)


marcoscaceres:
Yep, "const-type" should be fine.

marcoscaceres:
I think maybe let's do these incrementally, and once we have enough we can do a release.

PR #132, opened 2018-01-29
feat: add const-type for idlTypes 

saschanaz:
Now for `const-type`.

I'm not very happy with a field `idlType` again have its own `idlType`, can we rename it so that we can reduce same names?

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -445,7 +445,7 @@
         typ = consume(ID) || error("No type for const");
         typ = typ.value;
       }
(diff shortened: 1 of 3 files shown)


marcoscaceres:
> I'm not very happy with a field idlType again have its own idlType, can we rename it so that we can reduce same names?

I'm happy to go with whatever you think is best.

saschanaz:
Hmm. I have no good idea yet, I'll keep thinking about it.

PR #134, opened 2018-01-29
feat: add argument/return type

saschanaz:
Doing `return-type` will probably be enough rather than doing `operation-type`/`callback-type`, what do you think?


marcoscaceres:
> Doing return-type will probably be enough rather than doing operation-type/callback-type, what do you think?

I like `return-type` 👍 Good idea.

PR #135, opened 2018-01-30
Type on union idlType

saschanaz:
I think this will better fit for #93, as otherwise we have to go one level deeper to get the type of a union type. What do you think?

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -262,13 +262,13 @@
     function union_type(typeName) {
       all_ws();
       if (!consume(OTHER, "(")) return;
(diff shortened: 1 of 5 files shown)


marcoscaceres:
I guess this can work, worst case, when I hit a null, I just walk up one level to figure out what it is.

saschanaz:
Do you have a use case for detecting types of each child idlTypes in a union type?

marcoscaceres:
Probably not, as they are styled and linked to definitions either in the spec, or to another spec (and I know that the context is "this is an IDL Type of some sort, not a concept"). However, I'll have a play with what we have thus far and can let you know soon.

saschanaz:
Thanks! We'll see what happens.

PR #137, opened 2018-02-04
refactor: small syntax changes

saschanaz:
Keywords can be used as property names in ES5 so this PR removes quotes for them.

Originally I intended to use `array.join()` for string concatenation in writer.js and then found that [it's actually slower on V8](http://jsben.ch/ltdI2). Hmm.

```diff
 function iterate(things) {
   if (!things) return;
-  let ret = "";
-  for (const thing of things) ret += dispatch(thing);
-  return ret;
+  return things.map(dispatch).join('');
 };

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -58,7 +58,7 @@
       getter: false,
       setter: false,
       deleter: false,
-      "static": false,
+      static: false,
       stringifier: false
     });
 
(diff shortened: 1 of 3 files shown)


marcoscaceres:
Worth doing a release or should we sit on this for a bit?

saschanaz:
This doesn't boost performance or add any features, so we should sit on this.

PR #139, opened 2018-03-02
Optimise tokenisation and whitespace skipping

ricea:
Tokenisation and whitespace skipping are two major hot spots for the parser.

Optimise by:
1. Using re.exec() instead of str.replace()
2. Using re.lastIndex with sticky expressions to avoid repeatedly trimming the front of the string
3. Using str.indexOf('\n') when counting newlines
5. Using a zero-width lookahead to quickly reject integers in the "float" regexp
6. Tweaking other regexps
7. Peeking at the next character in tokenise() to reduce the number of regular expressions that need to be checked at each position
8. Moving initialisation of re tables out of the hot path

Also optimise the removal of quotation marks from strings by using str.substring() instead of str.replace().

Due to optimisation of the "whitespace" regexp tokenise() now splits "  // foo" into two tokens ("  " and "// foo") rather than one. This makes no difference to the output of the parser. Verified with the test suite and with the w-p-t "idlharness" tests.

With these changes tokenize() is 30% faster and all_ws() is no longer a major contributor to parse time.

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -188,6 +188,8 @@
           let lastIndex = 0;
           while (lastIndex < w.length) {
             let matched = false;
+            // Servo doesn't support using "const" in this construction yet.
+            // |type| can be made const once Servo supports it.
             for (let type in all_ws_re) {
               const re = all_ws_re[type];
               re.lastIndex = lastIndex;
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -189,6 +189,7 @@
           while (lastIndex < w.length) {
             let matched = false;
             // Servo doesn't support using "const" in this construction yet.
+            // See https://github.com/servo/servo/issues/20231.
             // |type| can be made const once Servo supports it.
             for (let type in all_ws_re) {
               const re = all_ws_re[type];
@@ -489,10 +490,6 @@
           return { type: "sequence", value: [] };
         } else {
           const str = consume(STR) || error("No value for default");
-          if (!str.value.startsWith('"'))
-            error(`string '${str.value}' doesn't start with a quote`);
-          if (!str.value.endsWith('"'))
-            error(`string '${str.value}' doesn't end with a quote`);
           str.value = str.value.slice(1, -1);
           return str;
         }
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -178,7 +178,7 @@
           let lastIndex = 0;
           while (lastIndex < w.length) {
             let matched = false;
-            for (var type in all_ws_re) {
+            for (let type in all_ws_re) {
               const re = all_ws_re[type];
               re.lastIndex = lastIndex;
               const result = re.exec(w);
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -45,7 +45,7 @@
       } else if (/[A-Z_a-z]/.test(nextChar)) {
         result = attemptTokenMatch(str, "identifier", tokenRe.identifier,
                                    lastIndex, tokens);
-      } else if (/"/.test(nextChar)) {
+      } else if (nextChar === '"') {
         result = attemptTokenMatch(str, "string", tokenRe.string,
                                    lastIndex, tokens);
         if (result === -1) {
@@ -137,24 +137,25 @@
       if (!tokens.length || tokens[0].type !== type) return;
       if (typeof value === "undefined" || tokens[0].value === value) {
         last_token = tokens.shift();
-        if (type === ID && last_token.value.charAt(0) === '_')
+        if (type === ID && last_token.value.startsWith('_'))
           last_token.value = last_token.value.substring(1);
         return last_token;
       }
+    }
+
+    function count(str, char) {
+      let total = 0;
+      for (let i = str.indexOf(char); i !== -1; i = str.indexOf(char, i + 1)) {
+        ++total;
+      }
+      return total;
     }
 
     function ws() {
       if (!tokens.length) return;
       if (tokens[0].type === "whitespace") {
         const t = tokens.shift();
-        let i = 0;
-        while (i !== -1) {
-          i = t.value.indexOf('\n', i);
-          if (i !== -1) {
-            ++line;
-            ++i;
-          }
-        }
+        line += count(t.value, '\n');
         return t;
       }
     }
@@ -476,11 +477,11 @@
           return { type: "sequence", value: [] };
         } else {
           const str = consume(STR) || error("No value for default");
-          if (str.value.charAt(0) !== '"')
+          if (!str.value.startsWith('"'))
             error(`string '${str.value}' doesn't start with a quote`);
-          if (str.value.charAt(str.value.length - 1) !== '"')
+          if (!str.value.endsWith('"'))
             error(`string '${str.value}' doesn't end with a quote`);
-          str.value = str.value.substring(1, str.value.length - 1);
+          str.value = str.value.slice(1, -1);
           return str;
         }
       }
@@ -968,7 +969,7 @@
           return ret;
         }
         const val = consume(STR) || error("Unexpected value in enum");
-        val.value = val.value.substring(1, val.value.length - 1);
+        val.value = val.value.slice(1, -1);
         ret.values.push(val);
         all_ws(store ? vals : null);
         if (consume(OTHER, ",")) {
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -1,29 +1,69 @@
 "use strict";
 
 (() => {
+  // These regular expressions use the sticky flag so they will only match at
+  // the current location (ie. the offset of lastIndex).
+  const tokenRe = {
+    // This expression uses a lookahead assertion to catch false matches
+    // against integers early.
+    "float": /-?(?=[0-9]*\.|[0-9]+[eE])(([0-9]+\.[0-9]*|[0-9]*\.[0-9]+)([Ee][-+]?[0-9]+)?|[0-9]+[Ee][-+]?[0-9]+)/y,
+    "integer": /-?(0([Xx][0-9A-Fa-f]+|[0-7]*)|[1-9][0-9]*)/y,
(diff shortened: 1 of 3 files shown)


marcoscaceres:
@dontcallmedom, seems TravisCI was disabled for this repo too? Any idea what’s been causing that?

dontcallmedom:
@marcoscaceres  no - we've had that happen a number of times the past few weeks, and we have no idea what's happening

marcoscaceres:
Pinged GitHub support. Will see if they can hopefully clue us in.

marcoscaceres:
@dontcallmedom, sorry, could you check if it’s correctly re-enabled? I’m unsure if I have enough privileges (I clicked on things, but just want to make sure).

dontcallmedom:
@marcoscaceres confirmed

marcoscaceres:
@ricea, could you send this again (or with an empty commit)? That should trigger TravisCI again. Want to confirm tests passing.

saschanaz:
That didn't trigger CI 🤔

lib/webidl2.js
```diff
@@ -97,7 +137,8 @@
       if (!tokens.length || tokens[0].type !== type) return;
       if (typeof value === "undefined" || tokens[0].value === value) {
         last_token = tokens.shift();
-        if (type === ID) last_token.value = last_token.value.replace(/^_/, "");
+        if (type === ID && last_token.value.charAt(0) === '_')
```
saschanaz: I would prefer `last_token.value.startsWith("_")`, IMO it shouldn't be slower than `charAt`...

lib/webidl2.js
```diff
@@ -429,7 +476,11 @@
           return { type: "sequence", value: [] };
         } else {
           const str = consume(STR) || error("No value for default");
-          str.value = str.value.replace(/^"/, "").replace(/"$/, "");
+          if (str.value.charAt(0) !== '"')
```
saschanaz: `startsWith` here too.

lib/webidl2.js
```diff
@@ -429,7 +476,11 @@
...
+            error(`string '${str.value}' doesn't start with a quote`);
+          if (str.value.charAt(str.value.length - 1) !== '"')
```
saschanaz: `endsWith` should be a cleaner alternative.

lib/webidl2.js
```diff
@@ -917,7 +968,7 @@
           return ret;
         }
         const val = consume(STR) || error("Unexpected value in enum");
-        val.value = val.value.replace(/"/g, "");
+        val.value = val.value.substring(1, val.value.length - 1);
```
saschanaz: `.slice(1, -1)` would be cleaner.

lib/webidl2.js
```diff
@@ -124,25 +174,22 @@
...
-          const wsTypes = [];
-          for (const k in re) wsTypes.push(k);
-          while (w.length) {
+          let lastIndex = 0;
+          while (lastIndex < w.length) {
             let matched = false;
-            for (const type of wsTypes) {
-              w = w.replace(re[type], (tok, m1) => {
-                store.push({ type: type + (pea ? ("-" + pea) : ""), value: m1 });
+            for (var type in all_ws_re) {
```
saschanaz: Are we going back to for-var-in pattern on purpose? I'm not sure we should...

lib/webidl2.js
```diff
@@ -106,14 +147,23 @@
...
+          }
+        }
         return t;
       }
     }
 
+    const all_ws_re = {
+      "ws": /([\t\n\r ]+)/y,
+      "line-comment": /\/\/(.*)\r?\n?/y,
+      "multiline-comment": /\/\*((?:[^*]|\*[^/])*)\*\//y
```
saschanaz: Curious about this multiline comment regex change, is there also a performance win?

lib/webidl2.js
```diff
@@ -1,29 +1,69 @@
 "use strict";
 
 (() => {
+  // These regular expressions use the sticky flag so they will only match at
+  // the current location (ie. the offset of lastIndex).
+  const tokenRe = {
+    // This expression uses a lookahead assertion to catch false matches
+    // against integers early.
+    "float": /-?(?=[0-9]*\.|[0-9]+[eE])(([0-9]+\.[0-9]*|[0-9]*\.[0-9]+)([Ee][-+]?[0-9]+)?|[0-9]+[Ee][-+]?[0-9]+)/y,
```
saschanaz: I want to note that this is now branching from [the spec](https://heycam.github.io/webidl/#idl-grammar). Maybe worth pushing to upstream if this gives a significant win.

lib/webidl2.js
```diff
@@ -1,29 +1,69 @@
...
+        }
+        if (result === -1) {
+          // '-' and '.' can also match "other".
+          result = attemptTokenMatch(str, "other", tokenRe.other,
+                                     lastIndex, tokens);
+        }
+      } else if (/[A-Z_a-z]/.test(nextChar)) {
+        result = attemptTokenMatch(str, "identifier", tokenRe.identifier,
+                                   lastIndex, tokens);
+      } else if (/"/.test(nextChar)) {
```
saschanaz: Just `nextChar === '"'`?

lib/webidl2.js
```diff
@@ -106,14 +147,23 @@
...
-          return m;
-        });
+        let i = 0;
+        while (i !== -1) {
+          i = t.value.indexOf('\n', i);
+          if (i !== -1) {
+            ++line;
+            ++i;
+          }
+        }
```
saschanaz: How about make this as `function count(str, char)`, for readability?

lib/webidl2.js
ricea: According to https://github.com/w3c/web-platform-tests/commit/280005efe04503f14891d73c6ceab1077a59a735 Servo does not support "const type of" yet. While we don't strictly need to care about it here, if we use it it will have to be patched downstream again.

lib/webidl2.js
ricea: I think it would be not be a good match for the spec because it makes the expression harder to read and understand. It offers a significant performance benefit in this tokeniser, because of the way we run the expression against every digit, most of which will fail the lookahead assertion. But it might just get in the way in a different type of tokeniser.

lib/webidl2.js
ricea: Yes, but not huge. It's about 2% faster on V8.

lib/webidl2.js
saschanaz: https://github.com/servo/servo/issues/19535

Servo supports `for (let type of foo)` (but in a wrong way), I think keeping for-of syntax should be preferred.

Alternatively the repo may just use a transpiled file. Should we provide one?

marcoscaceres: Maybe we should meet them half-way and use `let`... as there is only one `type` in scope, servo should not get confused, right?

ricea: I changed it to `let type in all_ws_re`. With the way the code is now it doesn't need `const ... of` anyway.

ricea:
@saschanaz I think that's everything. PTAL.

ricea:
I spotted a bug in my changes where a '/' that was not part of a comment would fail to be tokenised as "other". It doesn't make much difference at the moment because it's never valid IDL syntax, but it might in future. I added a regression test to ensure that the correct error is produced for a solitary '/'.

I took the opportunity to optimise tokenise() a bit more by treating whitespace as a separate case from comments (the parser still sees "whitespace").

lib/webidl2.js
saschanaz: I locally confirmed the latest nightly build has no issue with callback-less synchronous `for (let type of array)`, can we use it so that we can keep `of`?

saschanaz: Additionally, a relevant comment will be great so that we won't accidentally do a rollback.

ricea: We can't use `of` here unless we cache an array of the keys of `all_ws_re`. Is there a benefit in doing that?

I have added a comment about why we are not using `const`.

saschanaz: Ah, I just found that `wsTypes` is removed. Never mind, sorry.

I'd like to add a relevant Servo issue URL so that a future contributor can easily follow it, but I couldn't find one. Should I file a new issue?

ricea: If you have a running copy of Servo to repro on, that would be helpful, yes.

lib/webidl2.js
```diff
@@ -429,7 +489,11 @@
           return { type: "sequence", value: [] };
         } else {
           const str = consume(STR) || error("No value for default");
-          str.value = str.value.replace(/^"/, "").replace(/"$/, "");
+          if (!str.value.startsWith('"'))
+            error(`string '${str.value}' doesn't start with a quote`);
+          if (!str.value.endsWith('"'))
+            error(`string '${str.value}' doesn't end with a quote`);
```
saschanaz: Hmm, this cannot happen as the regex forces them, right? We have AST tests, so probably no need to do runtime check here. What do you think? @marcoscaceres

ricea: I don't think so. I just left it there in case there was something I was missing.

ricea:
@saschanaz I mostly benchmarked against a memory sanitizer build, as that is where I was feeling the pain (see http://crbug.com/810963). I just tested with a normal optimised compile against the dom/interfaces.html web-platform-test (using the Blink layout test infrastructure). I got a median of 1.36s before the change and 1.26s after the change, which is a 7% improvement. The improvements under msan were larger, as it spends a larger portion of its time in the parser.

marcoscaceres:
As an aside, maybe we should add some performance regressions testing using something similar to what http://crbug.com/810963 is doing.

ricea:
TravisCI appears to be stuck. :disappointed:

marcoscaceres:
Yeah, stuck everywhere:( all w3c projects currently affected. Hopefully will come back soon.

marcoscaceres:
Restarted Travis build... 🤞

marcoscaceres:
+ webidl2@10.2.1

PR #141, opened 2018-03-17
Always add extAttrs for .idlType

saschanaz:
1. An IDL Type can have extended attributes except when it's a return type, but currently it's not in the documentation.
2. `extAttrs` property for IDL types is not defined for types that don't have any extended attributes, and this is inconsistent with behaviors in e.g. arguments.

Thus this PR always adds `extAttrs` so that users don't have to check its existence.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,12 +93,13 @@ attached to a field called `idlType`:
 
 ```JS
 {
-  "type": "return-type",
+  "type": "attribute-type",
   "sequence": false,
   "generic": null,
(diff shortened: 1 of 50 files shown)


marcoscaceres:
@saschanaz, not sure if mentioned, but I granted you publish rights on npm. Feel free to publish, otherwise I’ll send it tomorrow or Monday.

saschanaz:
That's great, thanks!

PR #142, opened 2018-03-17
const-type as full IDL Type

saschanaz:
#141 didn't cover `const-type` so this PR adds full IDL type members to it. Those will always have the same values but it's better than forcing existence check.

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,7 +220,8 @@ A callback looks like this:
     "generic": null,
     "nullable": false,
     "union": false,
-    "idlType": "void"
+    "idlType": "void",
+    "extAttrs": []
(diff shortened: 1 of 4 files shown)


saschanaz:
I tried fixing https://github.com/w3c/respec/issues/1574 and then found this 😅

saschanaz:
10.3.1!

marcoscaceres:
Hmmm. Are tests not running on TravisCI again?

marcoscaceres:
Help with that ReSpec bug would be greatly appreciated. I started on it on Friday, but didn’t get enough time to figure it out.

PR #145, opened 2018-04-07
Fix whitespace handling of implements and includes.

kenrussell:
The all_ws() production here might return undefined for certain badly
formed WebIDL (for example, the simple text "foobar;"). Be careful not
to unshift an undefined value onto the tokens array.

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -1021,7 +1021,8 @@
         return ret;
       } else {
         // rollback
-        tokens.unshift(w);
+        if (w)
+          tokens.unshift(w);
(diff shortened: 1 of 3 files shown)


kenrussell:
Added a test and confirmed that it catches both bugs in whitespace unshifting. Please take another look. Thanks.

marcoscaceres:
@saschanaz feel free to merge and release. Otherwise, I can do it tomorrow.

saschanaz:
v10.3.2

marcoscaceres:
Awesome! thank you!

PR #148, opened 2018-05-06
Refactor simple_extended_attr();

saschanaz:
This PR:

1. avoids checking `(` twice in different if-branches
2. removes old MapClass related variable and comment
3. throws if an identifier list is empty

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -424,7 +424,10 @@
           consume(FLOAT) ||
           consume(INT) ||
           consume(STR);
-        if (!ret.rhs && consume(OTHER, "(")) {
+      }
+      all_ws();
+      if (consume(OTHER, "(")) {
+        if (eq && !ret.rhs) {
           // [Exposed=(Window,Worker)]
           const rhs_list = [];
           const id = consume(ID);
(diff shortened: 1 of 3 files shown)


lib/webidl2.js
```diff
@@ -671,13 +660,18 @@
         error("Unterminated stringifier");
     }
 
-    function identifiers(arr) {
+    function* identifiers() {
```
marcoscaceres: Admittedly, I'm not a huge fan of using generators for finite lists where values are not consumed incrementally. If the consumer of the generator is always going to be an array spreading the iterator, then we might as well leave this as returning an array. 

However, I'm open to discussion on the rationale for why you'd like this integrate this change.

saschanaz: I did this because generators can reduce a slight amount of boilerplate:

```js
function array() {
  const arr = [];
  for (;;) {
    arr.push(value);
  }
  return arr;
}

// vs 

function* generate() {
  for (;;) {
    yield value;
  }
}
```

But yeah, it's not a big deal here.

marcoscaceres: If it's ok, I'd prefer just having the `const arr =[];`. It lowers the cognitive overhead a little bit.

saschanaz: Done!

lib/webidl2.js
```diff
@@ -424,36 +424,24 @@
...
-        let args, pair;
-        // [Constructor(DOMString str)]
-        if (args = argument_list(store)) {
-          ret["arguments"] = args;
-        }
-        // [Constructor()]
         else {
-          ret["arguments"] = [];
+          // [NamedConstructor=Audio(DOMString src)] or [Constructor(DOMString str)]
+          ret.arguments = argument_list(store);
```
marcoscaceres: I wonder if there was some old version of node that got confused by this. No need to change it back :)

marcoscaceres:
about Travis being busted... do you have a link to where you saw that "repo not valid" error? I'm trying to get it working again.

marcoscaceres:
NM! found it! Reactivate it.... so weird.

marcoscaceres:
Still stuck tho.. not sure what to do. We can try sending some more empty commits, and see if that kicks it off again. I don't seem to have the option to restart the tests on the branch, as it's not showing up on travis. Any suggestions?

saschanaz:
Reopening worked!

marcoscaceres:
Oh, awesome! Feel free to merge the PRs and release.

PR #149, opened 2018-05-07
chore(package): bump version number to 10.3.3

saschanaz:
I basically did `npm version patch && npm publish && git push` and then I realized I cannot directly push into the main branch anymore. Maybe I should do version bump in PRs...

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "webidl2",
-  "version": "10.3.2",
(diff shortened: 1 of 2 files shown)


marcoscaceres:
Yeah, good point :/ I can grant you admin otherwise ... then you can push to develop when publishing, I think.

marcoscaceres:
Also, TIL: `npm version patch`... I had no idea.

Issue #125, opened 2017-12-17
Support full whitespace conservation


saschanaz:
Currently, e.g.

```
TypeA /* Hi, I'm a comment */ includes          TypeB
```

automatically becomes `TypeA includes TypeB` even with `ws: true`, as both parser and writer do not support full whitespace conservation.

marcoscaceres:
I ran into this too. It's an edge case but we should add it.

saschanaz:
Current plan (changes expected): introduce a new field named `trivia`.

1. An IDL definition has a `trivia` field that includes preceding whitespaces/comments of required tokens:

```json
{
  "trivia": {
    "base": "// The following is an interface type\n",
    "name": " /* I'm a comment between the word `interface` and the interface name */ ",
    "open": " ",
    "close": "\n"
  }
}
```

2. An optional feature is expressed as an object instead of boolean, again including a `trivia` field.

```
{
  "partial": {
    "trivia": " "
  },
  // "partial": null when not partial
}
```

3. Members also has its own `trivia` field.

marcoscaceres:
trivia seems fine/fun :)

PR #150, opened 2018-05-09
BREAKING CHANGE: remove opt.ws

saschanaz:
It has never been documented or thoroughly tested, and we are planning a replacement for this in #125.

I was trying some refactoring when trying #125 and this legacy blocked me.


lib/webidl2.js
```diff
@@ -174,37 +174,14 @@
       "line-comment": /\/\/(.*)\r?\n?/y,
       "multiline-comment": /\/\*((?:[^*]|\*[^/])*)\*\//y
     };
-    function all_ws(store, pea) { // pea == post extended attribute, tpea = same for types
+    function all_ws(pea) { // pea == post extended attribute, tpea = same for types
```
marcoscaceres: At some point, we should consider renaming these things properly. There is no reason to have such short argument and variable names. Will make this code much more maintainable and approachable.

saschanaz: Ooops, actually that `pea` is also should be removed as it's part of opt.ws.

saschanaz:
Should we add eslint to the CI so that we won't miss any unused variables?

marcoscaceres:
> Should we add eslint to the CI so that we won't miss any unused variables?

Yeah, I think that would be good. When reviewing, it's hard to know if a variable gets used or not, because often parts of a function gets hidden.

PR #151, opened 2018-05-09
Prevent incorrect enums

saschanaz:
The current implementation is more permissive than [the spec](https://heycam.github.io/webidl/#idl-enums) allows, so this PR fixes it:

```
Enum ::
    enum identifier { EnumValueList } ;
EnumValueList ::
    string EnumValueListComma
EnumValueListComma ::
    , EnumValueListString
    ε
EnumValueListString ::
    string EnumValueListComma
    ε
```

```
enum Empty {}; // shouldn't pass
enum NoComma { "value1" "value2" }; // shouldn't pass
```

(Should we mark this as a breaking change?)


marcoscaceres:
There are “bugs”, not breaking changes imo.

PR #68, opened 2017-05-17
Fix whitespace error on parsing extended attributes

SaschaNaz:
```WebIDL
[
  Constructor,
  Constructor(double radius)
] // Currently errors here because the parser does not consume whitespaces before consuming bracket
interface Circle {
  attribute double r;
  attribute double cx;
  attribute double cy;
  readonly attribute double circumference;
};
```

PR #154, opened 2018-05-11
Remove all_ws() and gather trivia implicitly

saschanaz:
...to better align with [the spec](https://heycam.github.io/webidl/#idl-grammar):

>Implicitly, any number of whitespace and comment terminals are allowed between every other terminal in the input text being parsed.

Now:

1. We don't have to manually add `all_ws()` everywhere (so that we can prevent bugs like #68).
2. All trivia tokens are stored, no need to manually process it when rollback (to prevent bugs like #145).
3. The stored trivia tokens help implementing #125 😁

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -175,36 +175,12 @@
       return total;
     }
 
-    function ws() {
-      if (!tokens.length) return;
-      if (tokens[0].type === "whitespace" || tokens[0].type === "comment") {
-        const t = tokens.shift();
-        line += count(t.value, '\n');
-        return t;
-      }
-    }
-
-    function all_ws() {
-      const t = { type: "whitespace", value: "" };
-      while (true) {
-        const w = ws();
-        if (!w) break;
-        t.value += w.value;
-      }
-      if (t.value.length > 0) {
-        return t;
-      }
-    }
-
     function integer_type() {
       let ret = "";
-      all_ws();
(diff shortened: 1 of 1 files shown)


lib/webidl2.js
```diff
@@ -15,57 +15,62 @@
...
+        result = attemptTokenMatch("other");
       }
       if (result === -1) {
         throw new Error("Token stream not progressing");
       }
       lastIndex = result;
     }
     return tokens;
+
+    function attemptTokenMatch(type, noFlushTrivia) {
```
marcoscaceres: Nit: we should try to avoid introducing another boolean trap. Can you make the second argument an object?

lib/webidl2.js
```diff
@@ -146,12 +153,20 @@
...
         last_token = tokens.shift();
+        line += count(last_token.trivia, "\n");
         if (type === ID && last_token.value.startsWith('_'))
           last_token.value = last_token.value.substring(1);
         return last_token;
       }
     }
 
+    function unconsume(...toks) {
+      for (let tok of toks) {
```
marcoscaceres: Nit: const tok

saschanaz: Oops, I forgot adding a comment, this was done because of Servo compatibility (was introduced in #139 and removed in #150).

marcoscaceres: ah yeah. Ok, cool.

PR #155, opened 2018-05-11
BREAKING CHANGE: remove allowNestedTypedefs

### LINKED ISSUE #104: Keep supporting allowNestedTypedefs or not?
We already issued several breaking changes, users probably removed the nonstandard syntax by now, so I think we can just remove it. What do you think?

saschanaz:
Closes #104 

Interestingly the corresponding test has already been removed by 1b5b83438e6a6220e638f38425e8910ceadd714a.

Hopefully this won't break anything 😁

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,23 +45,6 @@ In the browser:
 </script>
 ```
 
-### Advanced Parsing
-
-`parse()` can optionally accept a second parameter, an options object, which can be used to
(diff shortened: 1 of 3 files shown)


saschanaz:
v12.0.0

marcoscaceres:
Major releases, like it ain' no thang! 😎

PR #157, opened 2018-05-16
Prevent keywords from being identifiers

### LINKED ISSUE #156: Named terminal symbols cannot be identifiers
```webidl
interface interface {}; // shouldn't parse
enum enum {}; // shouldn't parse

interface _interface {}; // okay
```

Currently webidl2.js parses all string symbols as identifiers, instead [it should distinguish named terminal symbols](https://heycam.github.io/webidl/#idl-grammar):

>If the longest possible match could match one of the above named terminal symbols or one of the other terminal symbols from the grammar, it must be tokenized as the latter. Thus, the input text “long” is tokenized as the quoted terminal symbol `long` rather than an __identifier__ called "long", and “.” is tokenized as the quoted terminal symbol `.` rather than an __other__.

saschanaz:
Closes #156 

This adds `escapedName` field for operations/attributes/arguments/dictionary fields so that writer.js can use raw escaped name.

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -130,7 +130,9 @@
     }
 
     toString() {
-      return `${this.message}, line ${this.line} (tokens: ${JSON.stringify(this.input)})\n${JSON.stringify(this.tokens, null, 4)}`;
+      const escapedInput = JSON.stringify(this.input);
+      const tokens = JSON.stringify(this.tokens, null, 4);
+      return `${this.message}, line ${this.line} (tokens: ${escapedInput})\n${tokens}`;
     }
   }
 
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -199,10 +199,15 @@
       if (typeof value === "undefined" || tokens[0].value === value) {
         last_token = tokens.shift();
         line += count(last_token.trivia, "\n");
-        if (type === ID && last_token.value.startsWith('_'))
-          last_token.value = last_token.value.substring(1);
         return last_token;
       }
+    }
+
+    function unescape(identifier) {
+      if (identifier.startsWith('_')) {
+        identifier = identifier.slice(1);
+      }
+      return identifier;
     }
 
     function unconsume(...toks) {
@@ -378,7 +383,8 @@
         unconsume(type_token);
         return;
       }
-      ret.name = name.value;
+      ret.name = unescape(name.value);
+      ret.escapedName = name.value;
       if (ret.optional) {
         const dflt = default_();
         if (typeof dflt !== "undefined") {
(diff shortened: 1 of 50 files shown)


lib/webidl2.js
```diff
@@ -82,7 +130,7 @@
     }
 
     toString() {
-      return `${this.message}, line ${this.line} (tokens: '${this.input}')\n${JSON.stringify(this.tokens, null, 4)}`;
+      return `${this.message}, line ${this.line} (tokens: ${JSON.stringify(this.input)})\n${JSON.stringify(this.tokens, null, 4)}`;
```
marcoscaceres: nit: line length.

lib/webidl2.js
```diff
@@ -8,13 +8,57 @@
...
     "integer": /-?(0([Xx][0-9A-Fa-f]+|[0-7]*)|[1-9][0-9]*)/y,
-    "identifier": /[A-Z_a-z][0-9A-Z_a-z-]*/y,
+    "identifier": /_?[A-Za-z][0-9A-Z_a-z-]*/y,
     "string": /"[^"]*"/y,
     "whitespace": /[\t\n\r ]+/y,
     "comment": /((\/(\/.*|\*([^*]|\*[^\/])*\*\/)[\t\n\r ]*)+)/y,
     "other": /[^\t\n\r 0-9A-Z_a-z]/y
   };
 
+  const namedTerminals = [
```
marcoscaceres: nit: maybe we should alpha sort these. Might make it easier to maintain and spot duplicates, etc.

lib/webidl2.js
```diff
@@ -154,12 +199,17 @@
...
         last_token = tokens.shift();
         line += count(last_token.trivia, "\n");
-        if (type === ID && last_token.value.startsWith('_'))
-          last_token.value = last_token.value.substring(1);
         return last_token;
       }
     }
 
+    function unescape(identifier) {
+      if (identifier.startsWith('_')) {
```
marcoscaceres: Nit: maybe just?:
```JS
return identifier.startsWith('_') ? identifier.slice(1) : identifier; 
```

then there is no reassignment. But no problem to keep it as is.

lib/webidl2.js
saschanaz: I'm not sure what "line length" mean here... This is done to escape linebreaks:

```js
console.log(JSON.stringify("abc\nbcd"));
// "abc\nbcd" instead of:
// abc
// bcd

Comment (saschanaz) [reply lib/webidl2.js]: Maybe better to do `.replace()` rather than JSON.stringify..

Comment (marcoscaceres) [reply lib/webidl2.js]: sorry, I meant that this line of code is getting very long :) 
Could be: 

```JS
const input = JSON.stringify(this.input);
const tokens = JSON.stringify(this.tokens, null, 4);
return `${this.message}, line ${this.line} (tokens: ${input})\n${token}`;
```

saschanaz: Ah, okay, thanks!

saschanaz:
>Although I think the use of escaped name will be rare

I had to add it because one of the writer tests fail:

```webidl
// input
interface System {
  object createObject(DOMString _interface);
};

// previous output
interface System {
  object createObject(DOMString interface); // this now throws
};
```

marcoscaceres:
That's a good one, actually :)

saschanaz:
v12.1.0 (because `escapedName`)

Alexkho82:
#157

PR #161, opened 2018-05-17
fix: prevent tokeniser error when it meets _0

saschanaz:
It turns out I was not fully understanding the tokenizer error. The `other` regex was different from the one in the spec and it was causing token stream error. Now `_0` is parsed as: `_` as `other` and `0` as `integer`.

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -15,7 +15,7 @@
     "other": /[^\t\n\r 0-9A-Za-z]/y
   };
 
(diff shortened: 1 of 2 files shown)


saschanaz:
v12.1.2 (I had to jump v12.1.1 because of a73d090a31f8646f766d37d6a8115bbf01f73fd7. Filed https://github.com/npm/npm/issues/20650, sorry for any inconvenience.)

PR #163, opened 2018-05-20
fix: prevent incorrect union types

saschanaz:
https://heycam.github.io/webidl/#prod-UnionType

```
UnionType ::
    ( UnionMemberType or UnionMemberType UnionMemberTypes )
UnionMemberType ::
    ExtendedAttributeList NonAnyType
    UnionType Null
UnionMemberTypes ::
    or UnionMemberType UnionMemberTypes
    ε
```

```webidl
typedef (OnlyOne) UnionOne; // shouldn't pass, at least two types are required inside a union
```

Also it turns out we don't have any tests for incorrect union types so I added some.


saschanaz:
12.1.3

PR #165, opened 2018-05-21
fix: prevent readwrite attributes on namespaces 

saschanaz:
https://heycam.github.io/webidl/#index-prod-NamespaceMember

```
NamespaceMember ::
    RegularOperation
    readonly AttributeRest
```

A namespace attribute must be readonly, but currently readwrite attributes are allowed. This PR fixes it.

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -168,7 +168,7 @@
     function error(str) {
       const maxTokens = 5;
       const tok = tokens.slice(0, maxTokens).map(t => t.trivia + t.value).join("");
-      // Count newlines preceding the actual errorneous token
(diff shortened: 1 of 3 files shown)


lib/webidl2.js
```diff
@@ -774,15 +772,15 @@
...
+        const mem = noninherited_attribute({ readonly: true }) ||
           regular_operation() ||
           error("Unknown member");
         mem.extAttrs = ea;
         ret.members.push(mem);
       }
     }
 
-    function noninherited_attribute(prefix) {
+    function noninherited_attribute({ readonly } = {}) {
```
marcoscaceres: maybe we should default readonly to false instead of undefined?

PR #170, opened 2018-05-24
fix: write inherit readonly attribute in proper order

saschanaz:
https://heycam.github.io/webidl/#prod-ReadWriteAttribute

```
ReadWriteAttribute ::
    inherit ReadOnly AttributeRest
    AttributeRest
```

`inherit readonly`, not `readonly inherit` 😁

This only affects the currently unexposed writer so IMO no need for a version bump.


marcoscaceres:
Not related to this pr, but super weird these inherited attributes...

PR #172, opened 2018-05-24
refactor: merge operation functions

saschanaz:
I tried modifying `operation()` to fix something and then found I have to modify both operation functions. This PR merges them so we can have a single operation function.


marcoscaceres:
I like the consolidation

PR #173, opened 2018-05-24
fix: prevent incorrect spaced negative infinity and variadic marks

### LINKED ISSUE #162: Spaced three-dot shouldn't recognized as a variadic mark
```webidl
interface Interface {
  void valid(object... args); // okay
  void invalid(object. . . args); // shouldn't pass
};
```

This is a regression from #154. 😱

saschanaz:
Closes #162 

This is similar to #154 but for punctuations.

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -37,7 +37,7 @@
     "stringifier",
     "typedef",
     "unrestricted"
-  ]
+  ];
 
   const nonRegexTerminals = [
     "Infinity",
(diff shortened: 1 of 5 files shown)


lib/webidl2.js
```diff
@@ -198,7 +222,7 @@
 
     let last_token = null;
 
-    function consume(type, value) {
+    function consume_single(type) {
```
marcoscaceres: Seems messy to have the `consume_single` and `consume(...candidates)`, `consume(...candidates)` should consume single without needing to call out to the consume_single (i.e., merge `consume_single()` into `consume()`'s for loop).

PR #177, opened 2018-05-28
fix: consume return type for promise subtype

### LINKED ISSUE #168: Error parsing generic type with Promise<void>
Parsing https://github.com/w3c/web-platform-tests/blob/master/interfaces/BackgroundSync.idl is broken by https://github.com/w3c/webidl2.js/commit/621d0b6c17c6f34cf6b0b5c0976abb069703d4eb. 

```
Got an error during or right after parsing `interface SyncManager`: Error parsing generic type Promise, line 11 (tokens: "void> register(DOMString")
[
    {
        "type": "void",
        "value": "void",
        "trivia": ""
    },
    {
        "type": "other",
        "value": ">",
        "trivia": ""
    },
    {
        "type": "identifier",
        "value": "register",
        "trivia": " "
    },
    {
        "type": "other",
        "value": "(",
        "trivia": ""
    },
    {
        "type": "identifier",
        "value": "DOMString",
        "trivia": ""
    }
]
```

It looks like the relevant IDL is:

```
[Exposed=(Window,Worker)]
interface SyncManager {
  Promise<void> register(DOMString tag);
  Promise<sequence<DOMString>> getTags();
};
```

saschanaz:
This PR:

1. fixes #168 
2. fixes a bug where `Promise<void>?` was allowed (See https://heycam.github.io/webidl/#prod-NonAnyType)
1. adds generic/string type terminals as keywords
2. splits generic type parser into a new function `generic()`.
3. `return_type()` now does not throw, instead the callers should manage errors  with proper messages

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -15,6 +15,12 @@
     "other": /[^\t\n\r 0-9A-Za-z]/y
   };
 
+  const stringTypes = [
+    "ByteString",
+    "DOMString",
+    "USVString"
+  ];
+
   const argumentNameKeywords = [
     "attribute",
     "callback",
@@ -40,8 +46,10 @@
   ];
 
   const nonRegexTerminals = [
+    "FrozenArray",
     "Infinity",
     "NaN",
+    "Promise",
     "boolean",
     "byte",
     "double",
@@ -56,11 +64,13 @@
     "optional",
     "or",
     "readonly",
+    "record",
+    "sequence",
     "short",
     "true",
     "unsigned",
     "void"
-  ].concat(argumentNameKeywords);
+  ].concat(argumentNameKeywords, stringTypes);
 
   const punctuations = [
     "(",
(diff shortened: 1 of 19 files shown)


lib/webidl2.js
```diff
@@ -302,41 +312,55 @@
...
+          break;
+        case "sequence":
+        case "FrozenArray":
+          ret.idlType = type_with_extended_attributes(typeName);
+          break;
+        case "record":
+          if (probe("[")) error("Record key cannot have extended attribute");
+          ret.idlType = [];
+          const keyType = consume(...stringTypes);
+          if (!keyType) error("Record key must be DOMString, USVString, or ByteString");
```
marcoscaceres: nit: 
```
`Record key must be one of: ${stringTypes.join(", ")}.`
```

Then we don't need to maintain it if a new type gets added.

lib/webidl2.js
```diff
@@ -302,41 +312,55 @@
...
+          ret.idlType.push(Object.assign({ type: typeName }, EMPTY_IDLTYPE, { idlType: keyType.value }));
+          consume(",") || error("No comma after record key type");
```
marcoscaceres: s/`error("No`/`error("Missing`

lib/webidl2.js
```diff
@@ -302,41 +312,55 @@
...
+          const valueType = type_with_extended_attributes(typeName) || error("Error parsing generic type record");
+          ret.idlType.push(valueType);
+          break;
+      }
+      if (!ret.idlType) error(`Error parsing generic type ${name.type}`);
+      consume(">") || error(`No closing bracket after ${name.type}`);
```
marcoscaceres: As above: No -> Missing

lib/webidl2.js
```diff
@@ -545,7 +569,7 @@
       const name = consume(ID) || error("No name for callback");
       ret = current = { type: "callback", name: sanitize_name(name.value, "callback") };
       consume("=") || error("No assignment in callback");
-      ret.idlType = return_type();
+      ret.idlType = return_type() || error("No return type");
```
marcoscaceres: As above.

lib/webidl2.js
```diff
@@ -611,7 +635,7 @@
         else if (consume("deleter")) ret.deleter = true;
         else break;
       }
-      ret.idlType = return_type();
+      ret.idlType = return_type() || error("No return type");
```
marcoscaceres: As above

saschanaz:
I didn't really like double colons in `Got an error during or right after parsing `interface Foo`: Record key must be one of: ByteString, DOMString, USVString` so I replaced the second one with `->`. Not sure it's best, though.

marcoscaceres:
How about instead of "must be one of ->".., "must be one of `ByteString`, `DOMString`, `USVString`"

marcoscaceres:
updated comment above

saschanaz:
How about just `Record key must be a string type`?

saschanaz:
v13.0.0 because of #175.

Issue #178, opened 2018-05-29
No opening bracket after sequence


csnardi:
https://github.com/web-platform-tests/wpt/blob/master/input-events/idlharness.html fails to parse with v13.0.0 with the following error:

```
Got an error during or right after parsing `partial interface InputEvent`: No opening bracket after sequence, line 4 (tokens: " getTargetRanges();\n}")
[
    {
        "type": "identifier",
        "value": "getTargetRanges",
        "trivia": " "
    },
    {
        "type": "(",
        "value": "(",
        "trivia": ""
    },
    {
        "type": ")",
        "value": ")",
        "trivia": ""
    },
    {
        "type": ";",
        "value": ";",
        "trivia": ""
    },
    {
        "type": "}",
        "value": "}",
        "trivia": "\n"
    }
]
```

Relevant snippet of the IDL:

```WebIDL
partial interface InputEvent {
    readonly attribute DOMString     inputType;
    readonly attribute DataTransfer? dataTransfer;
    sequence<StaticRange> getTargetRanges();
};
```

marcoscaceres:
I wasn't able to reproduce. Parses fine for me with v13. @saschanaz, can you confirm?

saschanaz:
I'd say that it's a bug on WPT's side.

```
var f = document.createRange().createContextualFragment(`<pre id='idl'>
partial interface InputEvent {
    readonly attribute DOMString     inputType;
    readonly attribute DataTransfer? dataTransfer;
    sequence<StaticRange> getTargetRanges();
};

partial dictionary InputEventInit {
    DOMString             inputType = "";
    DataTransfer?         dataTransfer = null;
    sequence<StaticRange> targetRanges = [];
};
</pre>`);
console.log(f.children[0].textContent);
/*
partial interface InputEvent {
    readonly attribute DOMString     inputType;
    readonly attribute DataTransfer? dataTransfer;
    sequence getTargetRanges();
};

partial dictionary InputEventInit {
    DOMString             inputType = "";
    DataTransfer?         dataTransfer = null;
    sequence targetRanges = [];
};
*/
```

The code probably escape `<` and `>` for HTML parser. webidl2.js previously allowed `sequence` without brackets but not anymore.

csnardi:
Ah, that would make sense. Closing this then.

PR #179, opened 2018-05-30
refactor: unconsume by position

saschanaz:
This PR:

1. removes use of `array.unshift()` 😁
2. removes any need to remember all tokens to unconsume, prevents bugs like #145 
3. enables, but not in this PR, potentially better error message by showing preceding tokens in situations like #178. Not sure how should the format be.
4. fixes strange token order in error messages:
   ```
   > require(".").parse("interface X { operation(optional sequence<abc>); }");
   Thrown: Got an error during or right after parsing `interface X`: Unterminated operation, line 1 (tokens: ">optional); }")
   ```
5. always add `default` field for arguments. Previously it was added only when the value exists.

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -203,10 +203,12 @@
 
     function error(str) {
       const maxTokens = 5;
-      const tok = tokens.slice(0, maxTokens).map(t => t.trivia + t.value).join("");
+      const tok = tokens
+        .slice(consume_position, consume_position + maxTokens)
+        .map(t => t.trivia + t.value).join("");
       // Count newlines preceding the actual erroneous token
       if (tokens.length) {
-        line += count(tokens[0].trivia, "\n");
+        line += count(tokens[consume_position].trivia, "\n");
       }
 
       let message;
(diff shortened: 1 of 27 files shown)


lib/webidl2.js
```diff
@@ -229,35 +231,32 @@
...
       return identifier.startsWith('_') ? identifier.slice(1) : identifier;
     }
 
-    function unconsume(...toks) {
-      // TODO: use const when Servo updates its JS engine
-      // https://github.com/servo/servo/issues/20231
-      for (let tok of toks) {
-        line -= count(tok.trivia, "\n");
+    function unconsume(position) {
+      while (consume_position > position) {
```
marcoscaceres: Trying to figure out if this can get stuck in a loop and if we should guard against that. I couldn’t come up with anything tho.

saschanaz: It should be safe unless `position` is `-Infinity`, which will not happen.

marcoscaceres: yeah, that's what I thought too... but just wanted to make sure :)

saschanaz: Thinking twice is great 😁

saschanaz:
v13.0.1

PR #184, opened 2018-06-02
refactor(lib/writer): remove args()

saschanaz:
IMO `map()`+`join()` is better for readability than a for i++ loop 😁

Also removes obsolete `creator` and `legacycaller` from writer.

### DIFF
--- a/lib/writer.js
+++ b/lib/writer.js
@@ -45,23 +45,14 @@
       if (arg.default) ret += ` = ${const_value(arg.default)}`;
       return ret;
     };
-    function args(its) {
-      let res = "";
-      for (let i = 0, n = its.length; i < n; i++) {
-        const arg = its[i];
-        res += argument(arg);
-        if (i < n - 1) res += ",";
-      }
-      return res;
-    };
     function make_ext_at(it) {
       context.unshift(it);
       let ret = it.name;
       if (it.rhs) {
-        if (it.rhs.type === "identifier-list") ret += `=(${it.rhs.value.join(',')})`;
+        if (it.rhs.type === "identifier-list") ret += `=(${it.rhs.value.join(",")})`;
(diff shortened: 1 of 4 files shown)


lib/writer.js
```diff
@@ -153,7 +144,7 @@
     };
     function callback(it) {
       const ret = extended_attributes(it.extAttrs);
-      return `${ret}callback ${it.name} = ${type(it.idlType)}(${args(it.arguments)});`;
+      return `${ret}callback ${it.name} = ${type(it.idlType)}(${it.arguments.map(argument).join(",")});`;
```
marcoscaceres: Nit: I wonder if we should consider splitting these up a little bit, because any more complexity is going to make these hard to read and maintain... could be:

```JS
function callback(it) {
  const { extAttrs, name, arguments:args, idlType } = it; // "it" could have a better name too
  const extendedAttrs = extended_attributes(extAttrs);
  const type = type(it.idlType);
  const arguments = args.map(argument).join(",");
`${extendedAttrs}callback ${name} = ${type}(${arguments});`
```

saschanaz: A triple 👍 for better readability.

saschanaz: But how should the names of local variables be?

1. `const extendedAttrs = extended_attributes(extAttrs);` is bad because camelCase and underscore is mixed
2. `const type = type(it.idlType);` has a name collision.
3. `const arguments = args.map(argument).join(",");` is invalid in strict mode

IMO these are not just simple mistakes but show it's hard to name them.

marcoscaceres: Agree, naming sucks :( Why this was just a nit. Let's go with this for now but maybe come back to it if it becomes a problem.

saschanaz: This definitely is a problem in #185, hmm.

PR #185, opened 2018-06-03
BREAKING CHANGE: support full whitespace conservation

### LINKED ISSUE #125: Support full whitespace conservation

saschanaz:
Yeah... so this fixes #125, but probably this is too big to review.

Breaking changes here:

1. Modifiers are now objects rather than Booleans, with `trivia` field.
2. `extAttrs` field now gives a nullable object with `items` and `trivia` field rather than an array.
3. Some fields on operation objects are now moved into `body`, because `stringify;` is a bodyless operation.
4. `name` field on operations now gives a nullable object with `trivia` field, as operations can be anonymous.

This should finally enable ReSpec to upgrade to the latest webidl2.js.


saschanaz:
Most commits are intentionally independent, so I may open PRs for each commit. Thoughts? @marcoscaceres

marcoscaceres:
> Most commits are intentionally independent, so I may open PRs for each commit. Thoughts?

PR for each would be good... can then land them in pieced.

lib/writer.js
```diff
@@ -13,16 +13,19 @@
...
+      if (it.union) ret += `${it.trivia.open}(${it.idlType.map(type).join("")}${it.trivia.close})`;
       else {
-        if (it.generic) ret += `${it.generic}<`;
-        if (Array.isArray(it.idlType)) ret += it.idlType.map(type).join(", ");
-        else ret += type(it.idlType);
-        if (it.generic) ret += ">";
+        ret += it.trivia.base;
+        if (it.generic) ret += `${it.generic.value}${it.generic.trivia.open}<`;
+        if (Array.isArray(it.idlType)) ret += it.idlType.map(type).join("");
+        else ret += typeof it.idlType === "string" ? it.baseName : type(it.idlType);
```
marcoscaceres: Can we make generics always be arrays or either 1 or 2 length? Then we don't need the type check.

marcoscaceres: Alternative, if it's not an array, cast it into an array?

lib/writer.js
```diff
@@ -33,142 +36,170 @@
...
       return ret;
     };
     function extended_attributes(eats) {
-      if (!eats || !eats.length) return "";
-      return `[${eats.map(make_ext_at).join(", ")}]`;
+      if (!eats) return "";
+      return `${eats.trivia.open}[${eats.items.map(make_ext_at).join("")}${eats.trivia.close}]`;
     };
 
     const modifiers = "getter setter deleter stringifier static".split(" ");
```
marcoscaceres: I know this is no in the change sent but:
```
 const modifiers = "getter setter deleter stringifier static".split(" ");
```

Is kinda silly. Let's just do:

```JS
 const modifiers = ["getter",  "setter", "deleter", "stringifier", "static"];
```

Also, if they are not used elsewhere, we can just inline the array.

lib/writer.js
```diff
@@ -33,142 +36,170 @@
...
     };
     function includes(it) {
       const ret = extended_attributes(it.extAttrs);
-      return `${ret}${it.target} includes ${it.includes};`;
+      return `${ret}${it.trivia.target}${it.target}${it.trivia.includes}includes${it.trivia.mixin}${it.includes}${it.trivia.termination};`;
     };
     function callback(it) {
       const ret = extended_attributes(it.extAttrs);
-      return `${ret}callback ${it.name} = ${type(it.idlType)}(${it.arguments.map(argument).join(",")});`;
+      return `${ret}${it.trivia.base}callback${it.trivia.name}${it.name}${it.trivia.assign}=${type(it.idlType)}${it.trivia.open}(${it.arguments.map(argument).join("")}${it.trivia.close})${it.trivia.termination};`;
```
marcoscaceres: Nit: line length on this is bit too long.

lib/writer.js
```diff
@@ -33,142 +36,170 @@
...
+        if (v.separator) {
+          ret += `${v.separator.trivia}${v.separator.value}`;
+        }
       }
-      return ret + "};";
+      return `${ret}${it.trivia.close}}${it.trivia.termination};`;
     };
     function iterable(it) {
-      return `iterable<${Array.isArray(it.idlType) ? it.idlType.map(type).join(", ") : type(it.idlType)}>;`;
+      return `${it.trivia.type}iterable${it.trivia.open}<${Array.isArray(it.idlType) ? it.idlType.map(type).join("") : type(it.idlType)}${it.trivia.close}>${it.trivia.termination};`;
```
marcoscaceres: as above about type, maybe just:
```JS
const idlType = Array.isArray(it.idlType) ? idlType : [it.idlType];
```

lib/writer.js
```diff
@@ -33,142 +36,170 @@
...
     };
     function legacyiterable(it) {
-      return `legacyiterable<${Array.isArray(it.idlType) ? it.idlType.map(type).join(", ") : type(it.idlType)}>;`;
+      return `${it.trivia.type}legacyiterable${it.trivia.open}<${Array.isArray(it.idlType) ? it.idlType.map(type).join("") : type(it.idlType)}${it.trivia.close}>${it.trivia.termination};`;
     };
     function maplike(it) {
-      return `${it.readonly ? "readonly " : ""}maplike<${it.idlType.map(type).join(", ")}>;`;
+      return `${it.readonly ? `${it.readonly.trivia}readonly` : ""}${it.trivia.type}maplike${it.trivia.open}<${it.idlType.map(type).join("")}${it.trivia.close}>${it.trivia.termination};`;
```
marcoscaceres: nit: line length getting a little long here too..

saschanaz:
@marcoscaceres, I think this PR is finally small enough to review 🎉

PR #186, opened 2018-06-04
Add trivia field for interfaces/mixins

saschanaz:
This is the very first commit of #185.

This adds `trivia` field for interfaces and mixins, with no breaking change.


saschanaz:
Should I publish a minor release for each PR, or a major one after all of #185 are merged?

marcoscaceres:
No, let’s do one big one. Might create too much noise otherwise.

PR #188, opened 2018-06-05
Generics always as an array

saschanaz:
A simpler logic.


saschanaz:
@marcoscaceres I pushed some changes requested in #185. That was quick! 👍😆

PR #189, opened 2018-06-05
BREAKING CHANGE: move nullable field from const type to .idlType

saschanaz:
It's weird that `cnt.nullable` is `true` and then `cnt.idlType.nullable` is `false`.

Actually more weird thing is that the const type is allowed to be nullable (https://github.com/heycam/webidl/issues/448).

Also some refactorings in writer.js.

### DIFF
--- a/lib/writer.js
+++ b/lib/writer.js
@@ -37,7 +37,7 @@
         if (it.rhs.type === "identifier-list") ret += `=(${it.rhs.value.join(",")})`;
         else ret += `=${it.rhs.value}`;
       }
-      if (it.arguments) ret += `(${it.arguments.length ? it.arguments.map(argument).join(",") : ""})`;
+      if (it.arguments) ret += `(${it.arguments.map(argument).join(",")})`;
       return ret;
     };
     function extended_attributes(eats) {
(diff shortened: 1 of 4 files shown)


lib/writer.js
```diff
@@ -127,27 +127,17 @@
...
-    };
-    function maplike(it) {
-      return `${it.readonly ? "readonly " : ""}maplike<${it.idlType.map(type).join(", ")}>;`;
+      const ext = extended_attributes(it.extAttrs);
+      const values = it.values.map(v => `"${v.value}",`).join("");
+      return `${ext}enum ${it.name} {${values}};`;
     };
-    function setlike(it) {
-      return `${it.readonly ? "readonly " : ""}setlike<${type(it.idlType[0])}>;`;
+    function iterable_like(it) {
```
marcoscaceres: 👍

marcoscaceres:
Are we jumping the shark here wrt heycam/webidl#448 ?

saschanaz:
This PR is about moving the nullable feature into `idlType`, not removing it. Previously even a nullable const member type has `.idlType.nullable === false`, which is fairly unexpected.

And yeah, sorry for the misleading title 😞.

test/syntax/json/nullable.json
```diff
@@ -6,11 +6,10 @@
         "members": [
             {
                 "type": "const",
-                "nullable": true,
                 "idlType": {
                     "type": "const-type",
                     "generic": null,
-                    "nullable": false,
+                    "nullable": true,
```
saschanaz: (This is the main point of this PR)

saschanaz:
I hope that could be an answer. This PR still conforms to the standard but in more consistent way.

PR #192, opened 2018-06-09
Add trivia for idlTypes

saschanaz:
These four commits are mutually related (so I couldn't fix writer.js for each commit), but still each commit has its own baseline changes.

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -402,7 +402,7 @@
         ret.trivia.base = name.trivia;
         if (probe("<")) error(`Unsupported generic type ${name.value}`);
       }
-      if (ret.generic === "Promise" && probe("?")) {
+      if (ret.generic && ret.generic.value === "Promise" && probe("?")) {
         error("Promise type cannot be nullable");
       }
       type_suffix(ret);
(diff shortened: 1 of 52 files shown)


lib/webidl2.js
```diff
@@ -195,9 +195,13 @@
 
     const EMPTY_IDLTYPE = Object.freeze({
       generic: null,
-      nullable: false,
+      nullable: null,
```
marcoscaceres: Can you help me understand why we don’t default to false? Null seems a little odd here.

lib/webidl2.js
```diff
@@ -307,7 +335,10 @@
     }
 
     function type_suffix(obj) {
-      obj.nullable = !!consume("?");
+      const nullable = consume("?");
+      if (nullable) {
+        obj.nullable = { trivia: nullable.trivia };
```
saschanaz: >Can you help me understand why we don’t default to false?

@marcoscaceres I had to determine where should information about optional "modifier" token go, and the result is this.

I could do `idlType.trivia.nullable === null`, but then we would have two fields about nullable token being absent. 

The `null` way keeps compatibility with `if (idlType.nullable)` and can also benefit from [future optional chaining syntax](https://github.com/tc39/proposal-optional-chaining/): `idlType.nullable?.trivia`.

marcoscaceres: Thanks for the clarification, makes sense.

PR #196, opened 2018-06-10
Use eslint

### LINKED ISSUE #167: Use ESLint (with Travis)
While pulling in new versions of webidl2.js to web platform tests and users of web platform tests (Chromium), I've noticed lines with trailing whitespace from time to time. Perhaps ESLint should be included 
and used so this doesn't occur?

saschanaz:
Config file generated by `npx eslint --init`.

Closes #167

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,320 @@
+{
+    "env": {
+        "browser": true,
+        "commonjs": true,
+        "amd": true,
+        "es6": true,
+        "node": true,
+        "mocha": true
+    },
+    "extends": "eslint:recommended",
+    "rules": {
+        "accessor-pairs": "error",
+        "array-bracket-newline": "error",
+        "array-bracket-spacing": [
+            "error",
+            "never"
+        ],
+        "array-callback-return": "error",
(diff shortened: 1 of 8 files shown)


.eslintrc.json
```diff
@@ -0,0 +1,316 @@
...
+    "env": {
+        "browser": true,
+        "commonjs": true,
+        "amd": true,
+        "es6": true,
+        "node": true,
+        "mocha": true
+    },
+    "extends": "eslint:recommended",
+    "rules": {
```
marcoscaceres: Given that we are extending “eslint:recommended”, it possible to only list the ones we are changing?

saschanaz: So it seems we can just remove those `off` lines, but IMO maybe those lines are helpful for future reconsideration (at least we know they exist!). Should they be removed?

marcoscaceres: I don’t mind keeping it as is, but I’m worried that we don’t know which are set by us or why - and the recommended set can change or have new things added. I guess it’s not a big deal as the code base is only small.

saschanaz:
Replaced by #202

PR #200, opened 2018-06-10
Add trivia for extended attribute identifiers

saschanaz:
The list now contains objects instead of strings.

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -768,16 +768,14 @@
 
     function identifiers() {
       const arr = [];
-      const id = consume(ID);
-      if (id) {
-        arr.push(id.value);
-      }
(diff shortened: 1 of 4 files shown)


lib/webidl2.js
```diff
@@ -768,16 +768,14 @@
...
-      if (id) {
-        arr.push(id.value);
-      }
-      else error("Expected identifiers but not found");
-      while (true) {
-        if (consume(",")) {
-          const name = consume(ID) || error("Trailing comma in identifiers list");
-          arr.push(name.value);
-        } else break;
+      const id = untyped_consume(ID) || error("Expected identifiers but not found");
```
marcoscaceres: Nit: “but none found”

PR #201, opened 2018-06-11
Add trivia for extended attributes

saschanaz:
1. The field `extAttrs` now gives an object with `items` and `trivia` field.
2. `arguments` list are now moved into the new field `signature` which also has `trivia` for parentheses.

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -454,7 +454,7 @@
     }
 
     function type_with_extended_attributes(typeName) {
-      const extAttrs = extended_attrs() || null;
+      const extAttrs = extended_attrs();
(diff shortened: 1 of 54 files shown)


lib/webidl2.js
```diff
@@ -454,16 +454,16 @@
     }
 
     function type_with_extended_attributes(typeName) {
-      const extAttrs = extended_attrs();
+      const extAttrs = extended_attrs() || null;
```
marcoscaceres: I wonder if extended_attrs() should just return null when appropriate?

saschanaz: I felt I shouldn't do that as no other functions return null, but yeah, I think that's better.

PR #205, opened 2018-06-13
Add trivia for partial types

saschanaz:
Renames a local variable `isPartial` as `partialModifier`.

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -665,7 +665,7 @@
       if (!consume("callback")) return;
       const tok = consume("interface");
       if (tok) {
-        ret = interface_rest(false, "callback interface");
+        ret = interface_rest(null, "callback interface");
         ret.trivia.base = tok.trivia;
         return ret;
       }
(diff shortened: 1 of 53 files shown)


lib/webidl2.js
```diff
@@ -851,7 +851,7 @@
       return ret;
     }
 
-    function interface_rest(isPartial, typeName = "interface") {
+    function interface_rest(partialModifier = null, typeName = "interface") {
```
marcoscaceres: I'd really prefer if this was:
```JS
interface_rest(typeName = "interface", { partialModifier = null } ) {
```
It kills the annoying partialModifier boolean/null trap.

lib/webidl2.js
```diff
@@ -889,7 +889,7 @@
       }
     }
 
-    function mixin_rest(isPartial) {
+    function mixin_rest(partialModifier = null) {
```
marcoscaceres: I guess same applies here.

lib/webidl2.js
```diff
@@ -927,17 +927,17 @@
       }
     }
 
-    function interface_(isPartial) {
+    function interface_(partialModifier = null) {
```
marcoscaceres: same... and for the others too.

PR #207, opened 2018-06-14
Add trivia for enums and typedefs

saschanaz:
Exposes `trivia` field for enum values.


lib/webidl2.js
```diff
@@ -1039,44 +1039,57 @@
...
       const vals = [];
       const ret = current = {
         type: "enum",
         name: sanitize_name(name.value, "enum"),
-        values: vals
+        values: vals,
+        trivia
       };
-      consume("{") || error("No curly for enum");
+      const open = consume("{") || error("No curly for enum");
```
marcoscaceres: “Curly” might be confusing. Maybe “no opening `{`...”

lib/webidl2.js
```diff
@@ -1039,44 +1039,57 @@
...
+      trivia.open = open.trivia;
       let value_expected = true;
       while (true) {
-        if (consume("}")) {
+        const close = consume("}");
```
marcoscaceres: Had expected to see “No closing } for enum” somewhere as an error. Is that possible?

saschanaz: Currently `enum X {` throws with *Unexpected value in enum*, which is weird here. I introduced `eof` type in #185, maybe we can do better with it by `Unexpected end of file`.

saschanaz: Filed #208, we may revisit it later.

PR #214, opened 2018-06-18
Document trivia for root type declarations/members

saschanaz:
Update for `idlType`/`extAttrs` will follow after this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -496,34 +496,47 @@ An operation looks like this:
 ```JS
 {
   "type": "operation",
-  "getter": false,
-  "setter": false,
-  "deleter": false,
-  "static": false,
-  "stringifier": false,
-  "idlType": {
-    "type": "return-type",
-    "generic": null,
-    "nullable": false,
-    "union": false,
-    "idlType": "void",
-    "extAttrs": []
-  },
-  "name": "intersection",
-  "arguments": [{
-    "optional": false,
-    "variadic": true,
-    "extAttrs": [],
+  "getter": null,
+  "setter": null,
+  "deleter": null,
+  "static": null,
+  "stringifier": null,
+  "body": {
     "idlType": {
-      "type": "argument-type",
+      "type": "return-type",
       "generic": null,
       "nullable": false,
       "union": false,
-      "idlType": "long",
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -158,7 +158,7 @@ The fields are as follows:
 
 * `type`: Always "interface".
 * `name`: The name of the interface.
-* `partial`: An object with a string type field `trivia` for `partial` interface modifier. `null` if the type is not a partial interface.
+* `partial`: An object with a string type field `trivia` if a partial interface.
```
marcoscaceres: Reads a bit strangely... maybe:

If a partial interface, an object with a string type field `trivia`. Otherwise, ....

README.md
```diff
@@ -204,7 +204,7 @@ The fields are as follows:
 
 * `type`: Always "interface mixin".
 * `name`: The name of the interface mixin.
-* `partial`: An object with a string type field `trivia` for `partial` interface modifier. `null` if the type is not a partial interface mixin.
+* `partial`: An object with a string type field `trivia` if a partial interface mixin.
```
marcoscaceres: As above, and globally...

README.md
```diff
@@ -440,50 +497,69 @@ An operation looks like this:
...
 
 The fields are as follows:
 
 * `type`: Always "operation".
-* `getter`: True if a getter operation.
-* `setter`: True if a setter operation.
-* `deleter`: True if a deleter operation.
-* `static`: True if a static operation.
-* `stringifier`: True if a stringifier operation.
+* `getter`: An object with a string type field `trivia` if a getter operation.
```
marcoscaceres: As above, "If x, then Y. Otherwise, `null`." Same for the ones below...

README.md
```diff
@@ -574,17 +669,25 @@ The arguments (e.g. for an operation) look like this:
...
+    }
   }]
 }
 ```
 
 The fields are as follows:
 
-* `optional`: True if the argument is optional.
-* `variadic`: True if the argument is variadic.
+* `optional`: An object with a string type field `trivia` if the argument is optional.
```
marcoscaceres: as above...

PR #217, opened 2018-06-19
Document changes for idlType and extAttrs

saschanaz:
I think this is the last PR about `trivia`. 🎉


saschanaz:
Released as v14.0.0

PR #229, opened 2018-08-22
fix: escape non-member top identifiers

### LINKED ISSUE #228: Some leading underscores are no longer removed
Web IDL spec says:

> For all of these constructs [[named definitions](https://heycam.github.io/webidl/#dfn-named-definition), attributes, typedefs, dictionary members, constants, and operations], the [identifier](https://heycam.github.io/webidl/#dfn-identifier) is the value of the [**identifier**](https://heycam.github.io/webidl/#prod-identifier) token with any leading U+005F LOW LINE ("\_") character (underscore) removed.

However, webidl2.js no longer seems to remove some underscores, from identifiers that did not need them to start with:

```js
> require('webidl2').parse('interface _Int {};')
[ { type: 'interface',
    name: '_Int',
    partial: null,
    members: [],
    trivia:
     { base: '', name: ' ', open: ' ', close: '', termination: '' },
    inheritance: null,
    extAttrs: null },
  { type: 'eof', trivia: '' } ]
```

saschanaz:
Fixes #228 

This adds `escapedName` field for top level definitions, and also correctly checks duplication:

```webidl
interface Hi {};
interface _Hi {}; // this is a duplicate!
```

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -237,7 +237,7 @@
     function sanitize_name(name, type) {
       const unescaped = unescape(name);
       if (names.has(unescaped)) {
-        error(`The name "${unescaped}" of type "${names.get(unescaped)}" is already seen`);
+        error(`The name "${unescaped}" of type "${names.get(unescaped)}" was already seen`);
       }
       names.set(unescaped, type);
       return unescaped;
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -235,11 +235,12 @@
     }
 
     function sanitize_name(name, type) {
-      if (names.has(name)) {
-        error(`The name "${name}" of type "${names.get(name)}" is already seen`);
+      const unescaped = unescape(name);
+      if (names.has(unescaped)) {
+        error(`The name "${unescaped}" of type "${names.get(unescaped)}" is already seen`);
       }
-      names.set(name, type);
-      return name;
+      names.set(unescaped, type);
+      return unescaped;
     }
 
     let consume_position = 0;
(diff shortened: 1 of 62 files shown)


test/invalid/json/duplicate-escaped.json
```diff
@@ -0,0 +1,4 @@
+{
+    "message": "Got an error during or right after parsing `interface Iroha`: The name \"Iroha\" of type \"interface\" is already seen",
```
marcoscaceres: nit: s/is/was

lib/webidl2.js
```diff
@@ -235,11 +235,12 @@
     }
 
     function sanitize_name(name, type) {
-      if (names.has(name)) {
-        error(`The name "${name}" of type "${names.get(name)}" is already seen`);
+      const unescaped = unescape(name);
+      if (names.has(unescaped)) {
+        error(`The name "${unescaped}" of type "${names.get(unescaped)}" is already seen`);
```
marcoscaceres: Sorry, nit below belongs here.

test/invalid/json/duplicate-escaped.json
saschanaz: That means is->was, right? What does `s/` part mean?

marcoscaceres: Ah sorry, bad habit `s/` meaning replace... something we do at the w3c :(

marcoscaceres: (I think "s/" is a vim thing)

saschanaz: Oh, TIL an interesting w3c syntax 😁

TimothyGu:
Thanks!

PR #232, opened 2018-10-11
fix: enum values should be of type enum-value

### LINKED ISSUE #231: Enum value type should be "enum-value" not "string"
So, although enums are strings in JS, the type is not representative of the data type... it should be "enum-value".

marcoscaceres:
* closes #231 

- [ ]  update docs once this is merged.

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -1079,6 +1079,7 @@
           error("No comma between enum values");
         }
         const val = consume(STR) || error("Unexpected value in enum");
+        val.type = "enum-value";
         val.value = val.value.slice(1, -1);
         ret.values.push(val);
         val.separator = untyped_consume(",") || null;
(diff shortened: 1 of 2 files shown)


marcoscaceres:
Was not sure if this is a breaking change or not

marcoscaceres:
it might be a breaking change... given:
https://github.com/w3c/respec/pull/1850/files#diff-e45f88cac27757d7ac69556862e6226bR515

And:
https://github.com/w3c/respec/pull/1850/files#diff-e45f88cac27757d7ac69556862e6226bR758

@saschanaz wdyt?

marcoscaceres:
Note to self: update docs once this is merged.

saschanaz:
I agree, this also changes the documented API, so a major change.

marcoscaceres:
updating the docs

PR #234, opened 2018-10-30
eslint: prevent eval()

tripu:


tripu:
> *&ldquo;But we haven't use eval?&rdquo;*

I think you do: Travis builds run `npm test`, your `npm test` is `npm run lint && mocha`, and your `npm run lint` is `eslint lib/*.js test/*.js`.

saschanaz:
Um... I mean AFAIK we haven't use JS `eval()` in this library.

tripu:
Sorry, I understood *eslint* (instead of *eval*).
Well, now you *know* for sure that you do not use `eval()` :)

tripu:
Yeah, this small tweak to the eslint config is no big deal, I agree.
I was just looking at existing eslint config on W3C repos, to recommend a reasonable &ldquo;standard&rdquo; on [our Node.js guide](https://w3c.github.io/nodejs/doc/).
I guess you wouldn't do `if (true)` either, yet you have a rule to prevent that. And so on. That's the point of linting, right?
I'll let you decide whether you prefer a minimalistic set of rules, or something more explicit :slightly_smiling_face:

marcoscaceres:
As we live in the W3C organization, we should abide by the W3C's guidelines :) You are right the linting rules are not really for core maintainers, but for other possible contributors.

saschanaz:
>I guess you wouldn't do if (true) either, yet you have a rule to prevent that.

We won't, but we actually did! 🤣 https://github.com/w3c/webidl2.js/pull/202/files#diff-1854738585af51c74964a8acc17096c6

PR #237, opened 2018-11-17
refactor(lib/writer): reduce conditional constructs

saschanaz:
Removing them enables shorter code 😎 and eases further refactoring 👍


marcoscaceres:
Much nicer, indeed!

PR #239, opened 2018-11-18
refactor(lib/writer): use interface_like()

saschanaz:
`namespace()` and `dictionary()` are so much like `interface_()` so this PR removes them.

### DIFF
--- a/lib/writer.js
+++ b/lib/writer.js
@@ -108,13 +108,15 @@
       return `${inh.trivia.colon}:${inh.trivia.name}${inh.name}`;
     }
 
-    function interface_like(it, type = it.type) {
-      let ret = extended_attributes(it.extAttrs);
-      ret += token(it.partial, "partial");
-      ret += `${it.trivia.base}${type}${it.trivia.name}${it.escapedName}`;
-      ret += inheritance(it.inheritance);
-      ret += `${it.trivia.open}{${iterate(it.members)}${it.trivia.close}}${it.trivia.termination};`;
-      return ret;
+    function container(type) {
+      return it => {
+        let ret = extended_attributes(it.extAttrs);
+        ret += token(it.partial, "partial");
+        ret += `${it.trivia.base}${type}${it.trivia.name}${it.escapedName}`;
+        ret += inheritance(it.inheritance);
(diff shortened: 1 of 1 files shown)


lib/writer.js
```diff
@@ -108,10 +108,10 @@
       return `${inh.trivia.colon}:${inh.trivia.name}${inh.name}`;
     }
 
-    function interface_(it) {
+    function interface_like(it, type = it.type) {
```
marcoscaceres: I'm not a huge fan of taking the default value from the outer scope, tbh. I think we should just make the second argument mandatory. 

I wonder if we should actually call this "container" instead of "interface_like"... though it's true that they are "interface like", the thing that makes this the same is that they are the main container types.

lib/writer.js
```diff
@@ -182,19 +166,19 @@
       return `${readonly}${it.trivia.base}${it.type}${bracket}${it.trivia.termination};`;
     }
     function callbackInterface(it) {
-      return `${it.trivia.callback}callback${interface_(it)}`;
+      return `${it.trivia.callback}callback${interface_like(it, "interface")}`;
```
marcoscaceres: Could this be?
`${it.trivia.callback}${interface_like(it, "callback interface")}`;

lib/writer.js
```diff
@@ -182,19 +166,19 @@
...
     }
     function eof(it) {
       return it.trivia;
     }
 
     const table = {
-      interface: interface_,
+      interface: interface_like,
```
marcoscaceres: We could even get a little fancier here if we curry the `container` function:

```js
function containerMaker(type){
   return it => {

   }
}

const table = {
    interface: container("interface"),
    namespace: container("namespace") ,
    dictionary: container("dictionary"),
}

```

marcoscaceres: Fixed typo above

lib/writer.js
saschanaz: That will block emitting this as it is: `callback /* Undesirable but possible comment */ interface {};`

marcoscaceres: ah, right.

PR #240, opened 2018-11-24
BREAKING CHANGE: merge modifier fields

saschanaz:
The IDL spec currently allows only one special modifier:

```
ReadWriteAttribute ::
    inherit ReadOnly AttributeRest
    AttributeRest

SpecialOperation ::
    Special RegularOperation

Special ::
    getter
    setter
    deleter

StaticMember ::
    static StaticMemberRest

StaticMemberRest ::
    ReadOnly AttributeRest
    RegularOperation

Stringifier ::
    stringifier StringifierRest

StringifierRest ::
    ReadOnly AttributeRest
    RegularOperation
    ;
```

No attribute or operation can get multiple special modifiers, so we can now merge them into one single field and remove the need to check one-by-one.

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -192,11 +192,7 @@
 
     const EMPTY_OPERATION = Object.freeze({
       type: "operation",
-      getter: null,
-      setter: null,
-      deleter: null,
-      static: null,
-      stringifier: null,
+      special: null,
       body: null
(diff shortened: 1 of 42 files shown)


saschanaz:
https://github.com/heycam/webidl/pull/562 allows this 👍

PR #242, opened 2018-12-01
Add tests for writer template feature

saschanaz:
Sorry, my git client somehow pushed my branched commits into the develop branch 😱 I reverted them, and this is the PR.


saschanaz:
Starting to think some more tests are needed, e.g. call counting.

saschanaz:
Will open a new PR with unit tests rather than baseline comparisons.

PR #245, opened 2018-12-02
Add unit tests for writer template functions

saschanaz:


saschanaz:
Template functions shouldn't break anything, so 16.1.0 🎉

PR #253, opened 2018-12-09
fix: unescape includes/idlTypes/inheritances

### LINKED ISSUE #249: Include statements do not unescape type names
```webidl
_Foo includes _Bar;
```

The `target` and `includes` value should be `Foo` and `Bar`, but instead they are `_Foo` and `_Bar`.

Pretty an edge case, though.

### LINKED ISSUE #252: idlTypes should unescape type name
```js
const typedef = `typedef sequence<_Type> Baz;`
const [first] = require("webidl2").parse(typedef)[0].idlType.idlType;
first.idlType // should be `Type` but `_Type`
first.baseName // should be `Type` but `_Type`
```

Also an edge case.

saschanaz:
I was trying webidl2 upgrade in jsdom/webidl2js and found that this has been broken.

Fixes #249, fixes #252


saschanaz:
v17.0.1

PR #260, opened 2019-02-01
fix(checker): avoid using innerText

saschanaz:
[It was a Chromium bug](https://bugs.chromium.org/p/chromium/issues/detail?id=705478) that `innerText` inserted newlines to textarea. Now `innerText` does not add newlines so we should use `textContent` instead.


marcoscaceres:
Makes sense.

Issue #138, opened 2018-02-14
Meta - conforming Web IDL validator


marcoscaceres:
This bug is for planning... 

It would be really interesting to have webidl2.js be a validating parser... that is, one that checks the conformance requirements from the WebIDL spec and either warns, blows up, and/or collects errors/warnings.

saschanaz:
This can also be covered: https://github.com/w3c/respec/issues/1184

kenrussell:
What are the project owners' thoughts on how to get started on a large project like this?

In the WebGL project there is a longstanding issue https://github.com/KhronosGroup/WebGL/issues/2216 where the IDL is invalid, but I couldn't find a Web IDL parser that validates the Editor's Draft version of the spec ( https://heycam.github.io/webidl/ ).

Personally I'd like to contribute to webidl2.js, as writing the validation rules in [XML](https://github.com/dontcallmedom/webidl-checker/blob/master/widlproc-schematron/webidl-schematron.xml) doesn't sound fun.

Can you provide some ideas or framework under which validation rules can be contributed in a scalable manner by multiple individuals?

saschanaz:
IMO the implementation will be in JS with corresponding tests, with API like: `parse(text, { validate: true })`. The option `validate` should include the existing conformance check in #80, which then will be optional.

We have a very vague plan, and I think we should start with anything and then make some progress over it, including code readability optimization.

marcoscaceres:
Agree, but I'd like see some kind proposed structure for error reporting and done in an extensible way. I'd like us to consider: 
 
  * non-validating parser (basically what we have today, except for syntax errors).
  * validating parser (validates as it parsers, throws on errors? or reports/emits errors?).
  * output validator: runs non-validating parser, but then we can validate the output (i.e., `validate(parse(idl))`).

From ReSpec's perspective, the last option might be good, because it means we can still render parsed output, even if it's full of validation errors. 

The drawback of the final option is the loss of context (line numbers, for example). So, maybe we want something in the middle:

```JS
const idlObj = parse(text);
const { idl, errors } = parse(text, {validate: true));
```

`errors` would be validation errors that are non-fatal. We might need to change how the code currently works, because I think we throw a the moment for various things. 

There are tradeoffs for each. We would also need to refactor quite a bit of code, as the parser/validation aspects could lead to quite complex code, given the current code structure (i.e., we might need to split things up a bit more).

kenrussell:
[...]

PR #262, opened 2019-02-15
refactor: self-containing classes

saschanaz:
A trial to implement https://github.com/w3c/webidl2.js/issues/138#issuecomment-455918738.

### DIFF
--- a/lib/validator.js
+++ b/lib/validator.js
@@ -49,7 +49,7 @@
 
   function groupDefinitions(all) {
     const unique = new Map();
-    const duplications = new Set();
+    const duplicates = new Set();
     const partials = new Map();
     for (const def of all) {
       if (def.partial) {
@@ -67,14 +67,14 @@
       if (!unique.has(def.name)) {
         unique.set(def.name, def);
       } else {
-        duplications.add(def);
+        duplicates.add(def);
       }
     }
-    return { all, unique, partials, duplications };
+    return { all, unique, partials, duplicates };
   }
 
-  function* checkDuplicatedNames({ unique, duplications }) {
-    for (const dup of duplications) {
+  function* checkDuplicatedNames({ unique, duplicates }) {
+    for (const dup of duplicates) {
       const { name } = dup;
       const message = `The name "${name}" of type "${unique.get(name).type}" was already seen`;
       yield error(dup.source, dup.tokens.name, dup, message);
--- a/lib/validator.js
+++ b/lib/validator.js
@@ -3,7 +3,7 @@
 (() => {
   function error(source, token, current, str) {
     const consume_position = token.index;
-    const maxTokens = 5;
+    const maxTokens = 5; // arbitrary but works well enough
     const line =
       !source[token.index].type === "eof" ? source[consume_position].line :
       source.length > 1 ? source[consume_position - 1].line :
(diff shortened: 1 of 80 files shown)


marcoscaceres:
Looking good.

saschanaz:
No real value here yet, the commit here is for future self-containing class instances get their own token references that include position information.

TIL new GitHub draft PR so I wanted to try it 😆, IMO good as this will constantly remind me that I have a WIP branch.

marcoscaceres:
How does one do the Draft PR? That’s quite a useful feature.

saschanaz:
It's a fancy new feature: https://github.blog/2019-02-14-introducing-draft-pull-requests/

saschanaz:
IMO mostly done, not feel great about `toJSON()` but it's needed for JSON compatibility.

saschanaz:
Too much changes so I'll open smaller PRs soon.

marcoscaceres:
Appreciate that. Maybe we can do one class at a time? Generally ok if it's around ~100 per pull request.

saschanaz:
@marcoscaceres This still have two pieces of features: name duplication checker and a feature for https://github.com/KhronosGroup/WebGL/issues/2216. Should I split them?

marcoscaceres:
This is manageable now, so no need to split them. I'll try to review tonight or first thing tomorrow.

lib/validator.js
```diff
@@ -0,0 +1,159 @@
...
+
+    function lastLine(text) {
+      const splitted = text.split("\n");
+      return splitted[splitted.length - 1];
+    }
+  }
+
+  function groupDefinitions(all) {
+    const unique = new Map();
+    const duplications = new Set();
```
marcoscaceres: Nit: “duplicates”?

marcoscaceres: Actually, it’s fine to leave it. Would mean changing this in way too many places.

lib/validator.js
```diff
@@ -0,0 +1,159 @@
...
+      ...checkDuplicatedNames(defs),
+      ...checkInterfaceMemberDuplication(defs)
+    ];
+  }
+
+  const obj = {
+    validate
+  };
+
+  if (typeof module !== 'undefined' && typeof module.exports !== 'undefined') {
```
marcoscaceres: I wonder if we should just ship the ES6 Module and not UMD this? I guess we’d still need to bundle it for distribution but might be nice to ship just ES6.

lib/validator.js
saschanaz: That's why IDEs have variable renaming feature 😁

lib/validator.js
saschanaz: I'd like to do that, will open a PR adding webpack.

lib/validator.js
```diff
@@ -0,0 +1,159 @@
+"use strict";
+
+(() => {
+  function error(source, token, current, str) {
+    const consume_position = token.index;
+    const maxTokens = 5;
```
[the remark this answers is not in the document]
saschanaz: It's rather arbitrary. Hmm.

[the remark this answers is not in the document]
marcoscaceres: Yeah, that’s what I was wondering. It’s ok for it to be, but just sayin “it works well enough“ is fine.

PR #271, opened 2019-03-06
refactor: namespace as a class

saschanaz:
Namespace 👩‍🚀 this time.

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -966,6 +966,9 @@
         const { tokens } = instance;
         tokens.name = consume(ID) || error("No name for interface");
         current = instance;
+        if (!instance.partial) {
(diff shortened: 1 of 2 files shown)


lib/webidl2.js
```diff
@@ -1092,37 +1095,24 @@
...
         }
-        const ea = extended_attrs();
-        const mem = attribute({ noInherit: true, readonly: true }) ||
-          operation({ regular: true }) ||
-          error("Unknown member");
-        mem.extAttrs = ea;
-        ret.members.push(mem);
+        return Container.parse(new Namespace({ tokens }), {
+          type: "namespace",
+          allowedMembers: [
```
marcoscaceres: I like this means of laying out rules.

PR #272, opened 2019-03-06
refactor: interface/mixin as classes

saschanaz:
Baseline noise alert ⚠️

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -753,20 +753,16 @@
     }
 
     function callback() {
-      let ret;
-      const callbackToken = consume("callback");
-      if (!callbackToken) return;
+      const callback = consume("callback");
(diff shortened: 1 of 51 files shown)


lib/webidl2.js
```diff
@@ -1006,92 +1002,60 @@
...
-      while (true) {
-        const close = consume("}");
-        if (close) {
-          trivia.close = close.trivia;
-          const termination = consume(";") || error("Missing semicolon after interface");
-          trivia.termination = termination.trivia;
-          return ret;
+      get type() {
+        if (this.tokens.callback) {
+          return "callback interface";
```
marcoscaceres: Wish callback interfaces would go away entirely :( ... thankfully they are marked as legacy.

PR #273, opened 2019-03-06
refactor(lib/writer): use tokens object in container()

saschanaz:


lib/writer.js
```diff
@@ -180,20 +180,18 @@
...
-        const name = ts.name(it.escapedName, { data: it });
-        const head = wrap`${trivia.base}${type}${trivia.name}${name}`;
-        const inherit = inheritance(it.inheritance);
-        const body = brace_body(trivia, iterate(it.members, it));
-        return ts.definition(
-          wrap`${ext}${partial}${head}${inherit}${body}`,
-          { data: it }
-        );
-      };
+    function container(it) {
```
marcoscaceres: Maybe we can rename  “it” to something more specific?

saschanaz: Had no good idea. Do you have a suggestion? `data` would be helpful for `{ data }` but is not any specific.

marcoscaceres: No :( maybe will go through it after the refactor and think of something.

PR #280, opened 2019-03-14
BREAKING CHANGE: constant as a class

saschanaz:
This removes `valueLiteral` writer hook because it turns out that ReSpec doesn't need it.

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -407,29 +407,27 @@
     }
 
     function const_value() {
-      const token = consume("true", "false", "null", "Infinity", "-Infinity", "NaN", FLOAT, INT);
-      if (!token) {
-        return;
-      }
-      const { trivia } = token;
-      let data;
+      return consume("true", "false", "null", "Infinity", "-Infinity", "NaN", FLOAT, INT);
+    }
+
+    function const_data(token) {
       switch (token.type) {
         case "true":
         case "false":
-          data = { type: "boolean", value: token.type === "true" };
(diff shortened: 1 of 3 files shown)


lib/webidl2.js
```diff
@@ -407,29 +407,27 @@
     }
 
     function const_value() {
-      const token = consume("true", "false", "null", "Infinity", "-Infinity", "NaN", FLOAT, INT);
-      if (!token) {
-        return;
-      }
-      const { trivia } = token;
-      let data;
+      return consume("true", "false", "null", "Infinity", "-Infinity", "NaN", FLOAT, INT);
```
marcoscaceres: `FLOAT, INT` add indirection, why not just inline them? or do they have some special definition?

saschanaz: It's LONG and used twice. Not great to inline IMO.

marcoscaceres: Ok, I figured:) thought I’d ask.

saschanaz: Oops, maybe I misunderstood, I thought you asked about inlining `const_value()` which contains them.

For `FLOAT`/`INT`/`STR`/`ID`, they have slightly different meaning. Typically `consume("text")` means consuming a token which is exactly `text`, but in those cases regex is used instead.

marcoscaceres: ah, right. Sorry, I was reviewing on my phone and should have checked that. Using regexes 👍 makes sense then.

PR #283, opened 2019-03-15
refactor: extended attributes as a class

saschanaz:
The `list()` helper will be used in several more places.


lib/webidl2.js
```diff
@@ -326,6 +326,28 @@
       }
     }
 
+    function list({ parser, allowDangler, listName = "list" }) {
```
marcoscaceres: Maybe add a comment explaining what list covers?

PR #284, opened 2019-03-15
fix(lib/webidl2): disallow null on constants

saschanaz:
Following https://github.com/heycam/webidl/pull/686.

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -436,7 +436,7 @@
     }
 
     function const_value() {
-      return consume("true", "false", "null", "Infinity", "-Infinity", "NaN", FLOAT, INT);
+      return consume("true", "false", "Infinity", "-Infinity", "NaN", FLOAT, INT);
     }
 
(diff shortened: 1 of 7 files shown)


test/invalid/baseline/const-nullable.txt
```diff
@@ -0,0 +1,3 @@
+Syntax error at line 3, since `interface MyConstants`:
+  const boolean? ARE_WE_THERE_YET = false;
+               ^ No name for const
```
marcoscaceres: I wonder if we should identify this case... though I can't imagine anyone trying to do this.

lib/webidl2.js
```diff
@@ -772,6 +772,9 @@
           };
         }
         idlType = Object.assign({ type: "const-type" }, EMPTY_IDLTYPE, idlType);
+        if (probe("?")) {
+          error("Unexpected nullable constant type");
```
marcoscaceres: 😍

PR #289, opened 2019-03-16
refactor: generic type as a class

saschanaz:
Reaching to the end of this work 🚴‍♀️

### DIFF
--- a/lib/webidl2.js
+++ b/lib/webidl2.js
@@ -545,7 +545,7 @@
           }
           case "record": {
             if (probe("[")) error("Record key cannot have extended attribute");
-            const keyType = consume(...stringTypes) || error(`Record key must be a string type`);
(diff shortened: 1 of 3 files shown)


lib/webidl2.js
```diff
@@ -522,63 +522,59 @@
...
+          }
+          case "sequence":
+          case "FrozenArray": {
+            const subtype = type_with_extended_attributes(typeName) || error(`Missing ${base.type} subtype`);
+            ret.subtype.push(subtype);
+            break;
+          }
+          case "record": {
+            if (probe("[")) error("Record key cannot have extended attribute");
+            const keyType = consume(...stringTypes) || error(`Record key must be a string type`);
```
marcoscaceres: Only if not too annoying, might be nice: 
 
```suggestion
            const keyType = consume(...stringTypes) || error(`Record key must be one of: ${stringTypes.join(", ")}.`);
```

PR #291, opened 2019-03-18
refactor: remove remaining JSON idlTypes

saschanaz:
#262 will be finally reviewable after this.


saschanaz:
>Looking great! 💪

👍 🎉 , but... would you approve this then? 🤣

PR #293, opened 2019-03-20
refactor: use ES module syntax and bundle by webpack

saschanaz:
Nearly everything here is just whitespace changes.

PS: I hope Node.js support ES module by default soon, without its own new file extension.
PS2: Maybe we can split the parser now?


marcoscaceres:
> Maybe we can split the parser now?

Up to you. Would be nice as it’s own thing.

PR #295, opened 2019-03-21
refactor: use module for checker

saschanaz:
Because modules everywhere 😄

Also replacing default export with named exports, which will allow `import { parse } from "webidl2"`.

### DIFF
--- a/checker/index.html
+++ b/checker/index.html
@@ -2,31 +2,31 @@
 <html>
 
 <head>
-  <meta charset='utf-8'>
+  <meta charset="utf-8">
   <title>WebIDL 2 Checker</title>
 
-  <script type='module'>
-    import { parse, validate } from '../index.js';
+  <script type="module">
+    import { parse, validate } from "../index.js";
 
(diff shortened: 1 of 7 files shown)


checker/index.html
```diff
@@ -5,8 +5,9 @@
   <meta charset='utf-8'>
   <title>WebIDL 2 Checker</title>
 
-  <script src='../dist/webidl2.js'></script>
-  <script>
+  <script type='module'>
+    import { parse, validate } from '../index.js';
```
marcoscaceres: nit: could we settle on `"` for the file... we've got a mix of single and double quotes.

index.js
```diff
@@ -1,9 +1,3 @@
-import parse from "./lib/webidl2.js";
-import write from "./lib/writer.js";
-import validate from "./lib/validator.js";
-
-export default {
-  parse,
-  write,
-  validate
-};
+export { parse } from "./lib/webidl2.js";
```
marcoscaceres: oooh... never seen this syntax before! That's fancy!

PR #296, opened 2019-03-21
fix(lib/validator): validate when overloads are all from mixins

saschanaz:
Turns out the previous function didn't really cover https://github.com/KhronosGroup/WebGL/issues/2216.

Also adding `name` field for operations, which existed but later replaced by `body.name`. Existence check has been painful after that so this field should help.

### DIFF
--- a/lib/validator.js
+++ b/lib/validator.js
@@ -105,7 +105,7 @@ function* checkInterfaceMemberDuplication(defs) {
     for (const addition of additions) {
       const { name } = addition;
       if (name && existings.has(name)) {
-        const message = `The operation "${name}" has already been defined for the base interface "${base.name}"`;
+        const message = `The operation "${name}" has already been defined for the base interface "${base.name}" either in itself or in a mixin`;
         yield error(ext.source, addition.body.tokens.name, ext, message);
       }
     }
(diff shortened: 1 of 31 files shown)


lib/validator.js
```diff
@@ -89,32 +89,31 @@ function* checkInterfaceMemberDuplication(defs) {
...
-    for (const op of ext.members.filter(mem => mem.type === "operation")) {
-      const name = (op.body && op.body.name) ? op.body.name.value : "";
-      if (name && names.has(name)) {
-        const message = `The operation "${name}" has already been defined in the base interface "${base.name}"`;
-        yield error(ext.source, op.body.tokens.name, ext, message);
+  function* forEachExtension(additions, existings, ext, base) {
+    for (const addition of additions) {
+      const { name } = addition;
+      if (name && existings.has(name)) {
+        const message = `The operation "${name}" has already been defined for the base interface "${base.name}"`;
```
marcoscaceres: Don't know if it helps or too long... feel free to reject: 

```suggestion
        const message = `The operation "${name}" has already been defined for the base interface "${base.name} either in the base class or in a mixin"`;
```

saschanaz: Using the word `class` doesn't look good... Hmm.

PR #299, opened 2019-04-13
refactor: merge error functions

saschanaz:


lib/error.js
```diff
@@ -0,0 +1,63 @@
...
+  const procedingLine = procedingText.split("\n")[0];
+
+  const spaced = " ".repeat(precedingLine.length) + "^ " + message;
+  const contextualMessage = precedingLine + procedingLine + "\n" + spaced;
+
+  const contextType = type === "Syntax" ? "since" : "inside";
+  const grammaticalContext = current ? `, ${contextType} \`${current.partial ? "partial " : ""}${current.type} ${current.name}\`` : "";
+  return `${type} error at line ${line}${grammaticalContext}:\n${contextualMessage}`;
+
+  function sliceTokens(count) {
```
marcoscaceres: style nit: IMO, looks better if these function are hoisted above the actual code and then the last line in the overall function is the `return ...`.

saschanaz: I do prefer functions being underneath to allow function bodies to start with actual code, but it seems `parseByTokens()` in lib/webidl2 does return in the last line. I think I should follow the consistency way 😁

PR #300, opened 2019-04-17
feat: support constructor as a member

saschanaz:
For https://github.com/heycam/webidl/pull/700.


test/syntax/idl/constructor.widl
```diff
@@ -1,9 +1,9 @@
 // Extracted from http://dev.w3.org/2006/webapi/WebIDL/ on 2011-05-06
-[Constructor,
- Constructor(float radius)]
 interface Circle {
+  constructor();
+  constructor(float radius);
```
marcoscaceres: Wonder if we should add our own cases too with other argument things... probably not necessary given everything uses the same code path.

marcoscaceres: ..e.g. variadic, optional, defaults, etc.

saschanaz: I think it's not strictly needed but also does not harm. I'll add them as it's simple enough.

saschanaz:
>V

saschanaz:
>though should we sit on this till the WebIDL side merges?

https://github.com/heycam/webidl/pull/700#pullrequestreview-221042105 says the vise-versa thing, wouldn't it be a deadlock? I'm not sure.

marcoscaceres:
Ok, I think we should let them know we are ready on our side. We are a small piece of the solution: getting Gecko, Blink, etc. to update their binding layers will take some time. We could also add ReSpec support, which will also help move things along a little.

saschanaz:
Should be merged only when we are confident that browser implementors are on board.

marcoscaceres:
Yay! Looks like this can finally land... we should make a plan for what to do about specs that are using the old syntax.

saschanaz:
> Yay! Looks like this can finally land... we should make a plan for what to do about specs that are using the old syntax.

We can add an autofix, but I feel like I should wait until at least some of the spec actually start to use the new syntax.

marcoscaceres:
I'd like to use it for PaymentRequest :)

PR #304, opened 2019-04-18
BREAKING CHANGE: remove token fields

saschanaz:
They are now merged into `.tokens`.


saschanaz:
Poor trivia is finishing its move into `.tokens[tokenName].trivia`, to be not poor!

PR #315, opened 2019-04-30
docs(lib/writer): document generic()

saschanaz:
Forgot to update the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,12 +83,17 @@ var result = WebIDL2.write(tree, {
      */
     name: (escaped, { data, parent }) => escaped,
     /**
-     * Called for each type referece, e.g. `Window`, `Promise`, or `unsigned long`.
+     * Called for each type referece, e.g. `Window`, `DOMString`, or `unsigned long`.
      * @param escaped The referenced name. Typically string, but may also be the return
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -83,12 +83,17 @@ var result = WebIDL2.write(tree, {
...
      * @param escaped The referenced name. Typically string, but may also be the return
      *            value of `wrap()` if the name contains whitespace.
      * @param unescaped Unescaped reference.
      */
     reference: (escaped, unescaped) => escaped,
+    /**
+     * Called for each generic-form syntax, e.g. `sequence`, `Promise`, or `maplike`.
+     * @param {string} name The keyword for syntax
+     */
+    generic: name => name,
```
marcoscaceres: Just a minor thing... we don't need to fix now... but using arrow functions here is probably not ideal because `this` will weird for anyone copy/pasting and modifying this code. Like, if someone ends up building their own template object, and they want to do template.WHATEVER, then `this` will be wrong.

saschanaz: I don't expect people to use `this` in templates to refer the template object... Just didn't want to write full function body with braces.

PR #329, opened 2019-05-27
Remove trailing whitespace in lib/productions/helpers.js

foolip:


saschanaz:
Travis is broken and keeps being broken. Maybe we could migrate to GitHub Actions, thoughts? @marcoscaceres

saschanaz:
Merging because this passes local tests.

marcoscaceres:
I've not used GH actions, but if it does the same thing, then sure!

Issue #327, opened 2019-05-27
eof nodes


TimothyGu:
Since around ten major releases ago, webidl2 started emitting `eof` nodes:
```webidl
interface DOMImplementation {};
```
```json
[
  {
    "type": "interface",
    "name": "DOMImplementation",
    "escapedName": "DOMImplementation",
    "partial": null,
    "members": [],
    "trivia": {
      "base": "",
      "name": " ",
      "open": " ",
      "close": "",
      "termination": ""
    },
    "inheritance": null,
    "extAttrs": null
  },
  {
    "type": "eof",
    "value": "",
    "trivia": ""
  }
]
```

This is not documented anywhere, and arguably shouldn't even be in the AST as it doesn't contribute to the semantic meaning of the IDL fragment (it can be helpful for a CST though). Can one of the following happen:

- Document it
- Move it behind an option

Thanks

saschanaz:
Oops, didn't know it was undocumented, I'll add it.

I can add an option, maybe `{ concrete: true }`, but currently all it do would be removing `eof`. Maybe it's enough?

TimothyGu:
To be honest I don't mind if this will be the only one non-semantic node emitted by webidl2.js, now and in the future. What purpose did this intend to serve?

saschanaz:
It's mainly for our writer to keep the whole structure, for example:

```webidl
// This interface has a great feature!
interface X {};

// TODO: implement interface Y
// interface Y {};
```

Parsing this will generate an `eof` with the TODO comment, for our writer to be able to rewrite it intact: `webidl2.write(webidl2.parse(text))`

PR #331, opened 2019-05-28
docs(lib/webidl2): document `eof`

saschanaz:
Requested from #327. Not introducing an option yet.


saschanaz:
Published as 21.1.0

PR #332, opened 2019-06-02
fix(lib/webidl2): inherited attributes cannot be readonly

saschanaz:
Follows https://github.com/heycam/webidl/pull/704, but it has been already invalid.


saschanaz:
@marcoscaceres Seems we need to do the same Travis thing here. (Removing the service and marking the app as Required.)

marcoscaceres:
Done

saschanaz:
Yay, thanks!

PR #335, opened 2019-06-05
refactor(lib/webidl2): modularize Type

saschanaz:


lib/productions/type.js
```diff
@@ -0,0 +1,32 @@
...
+    super({ source, tokens });
+    Object.defineProperty(this, "subtype", { value: [] });
+    this.extAttrs = [];
+  }
+
+  get generic() {
+    return "";
+  }
+  get nullable() {
+    return !!this.tokens.nullable;
```
marcoscaceres: Can we do `Boolean(this.tokens.nullable)` instead? !! always confuses me.

PR #340, opened 2019-07-06
refactor(lib/webidl2): modularize attributes/operations

saschanaz:
Not much is remaining now 💪

### DIFF
--- a/lib/productions/attribute.js
+++ b/lib/productions/attribute.js
@@ -24,13 +24,13 @@ export class Attribute extends Base {
       tokeniser.unconsume(start_position);
       return;
     }
-    ret.idlType = type_with_extended_attributes(tokeniser, "attribute-type") || tokeniser.error("No type in attribute");
+    ret.idlType = type_with_extended_attributes(tokeniser, "attribute-type") || tokeniser.error("Attribute lacks a type");
     switch (ret.idlType.generic) {
       case "sequence":
       case "record": tokeniser.error(`Attributes cannot accept ${ret.idlType.generic} types`);
     }
-    tokens.name = tokeniser.consume("identifier", "required") || tokeniser.error("No name in attribute");
-    tokens.termination = tokeniser.consume(";") || tokeniser.error("Unterminated attribute");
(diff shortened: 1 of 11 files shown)


lib/productions/attribute.js
```diff
@@ -0,0 +1,52 @@
...
+    if (!tokens.base) {
+      tokeniser.unconsume(start_position);
+      return;
+    }
+    ret.idlType = type_with_extended_attributes(tokeniser, "attribute-type") || tokeniser.error("No type in attribute");
+    switch (ret.idlType.generic) {
+      case "sequence":
+      case "record": tokeniser.error(`Attributes cannot accept ${ret.idlType.generic} types`);
+    }
+    tokens.name = tokeniser.consume("identifier", "required") || tokeniser.error("No name in attribute");
```
marcoscaceres: I’m a bit unsure what “no name in attribute” means here. Maybe, “attribute lacks a name”?

lib/productions/attribute.js
```diff
@@ -0,0 +1,52 @@
...
+    tokens.termination = tokeniser.consume(";") || tokeniser.error("Unterminated attribute");
```
marcoscaceres: Maybe add, “, expected ;”

lib/productions/operation.js
```diff
@@ -0,0 +1,46 @@
...
+    }
+    if (!special && !regular) {
+      tokens.special = tokeniser.consume("getter", "setter", "deleter");
+    }
+    ret.idlType = return_type(tokeniser) || tokeniser.error("Missing return type");
+    tokens.name = tokeniser.consume("identifier");
+    tokens.open = tokeniser.consume("(") || tokeniser.error("Invalid operation");
+    ret.arguments = argument_list(tokeniser);
+    tokens.close = tokeniser.consume(")") || tokeniser.error("Unterminated operation");
+    tokens.termination = tokeniser.consume(";") || tokeniser.error("Unterminated attribute");
```
marcoscaceres: Unterminated operation as the error, no? Could be the same as above regarding the “;”.

saschanaz: Oh wow, never noticed that mistake. Good catch.

PR #343, opened 2019-07-07
refactor(lib/webidl2): modularize callback+iterable

saschanaz:
One more PR is left and I think the next one is about validating #338.

### DIFF
--- a/lib/productions/iterable.js
+++ b/lib/productions/iterable.js
@@ -1,0 +1,48 @@
+import { Base } from "./base";
+import { type_with_extended_attributes } from "./helpers";
+
+export class IterableLike extends Base {
+  /**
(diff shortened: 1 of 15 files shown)


lib/productions/iterable.js
```diff
@@ -0,0 +1,48 @@
...
+    if (!tokens.base) {
+      tokeniser.unconsume(start_position);
+      return;
+    }
+
+    const { type } = ret;
+    const secondTypeRequired = type === "maplike";
+    const secondTypeAllowed = secondTypeRequired || type === "iterable";
+
+    tokens.open = tokeniser.consume("<") || tokeniser.error(`Missing brackets in ${type} declaration`);
```
marcoscaceres: “Missing less-than sign (<)”, the word “bracket” is too ambiguous. Can do the same below for the closing greater than.

PR #346, opened 2019-07-08
refactor(lib/webidl2): modularize container types

saschanaz:
The last part of modularization.


saschanaz:
Yay, deeply thank you for your reviews on the whole modularization steps!

PR #347, opened 2019-07-08
feat: show filename when throwing

saschanaz:
For https://github.com/w3c/respec/issues/2235.


saschanaz:
Ah, I have to think more about the multifile cases...

saschanaz:
Just add support for array input. Might just be okay.

lib/error.js
```diff
@@ -50,10 +50,12 @@ function error(source, position, current, message, type) {
   const contextualMessage = precedingLine + sobsequentLine + "\n" + spaced;
 
   const contextType = type === "Syntax" ? "since" : "inside";
+  const fileContext = source.fileName ? ` in ${source.fileName}` : "";
```
marcoscaceres: Super annoying nit: filename is just one word in English 🥺

marcoscaceres:
```suggestion
  const fileContext = source.filename ? ` in ${source.filename}` : "";
```

lib/tokeniser.js
```diff
@@ -223,10 +223,11 @@ export class Tokeniser {
 }
 
 class WebIDLParseError extends Error {
-  constructor({ message, line, input, tokens }) {
+  constructor({ message, line, fileName, input, tokens }) {
```
marcoscaceres:
```suggestion
  constructor({ message, line, filename, input, tokens }) {
```

lib/tokeniser.js
```diff
@@ -223,10 +223,11 @@ export class Tokeniser {
...
     super(message);
     this.name = "WebIDLParseError"; // not to be mangled
     this.line = line;
+    this.fileName = fileName;
```
marcoscaceres:
```suggestion
    this.filename = filename;
```

lib/error.js
saschanaz: nit 2: it's not always for a file. What would be a better name? `trackingName`? Just `name`?

lib/webidl2.js
```diff
@@ -92,5 +92,8 @@ function parseByTokens(tokeniser, options) {
 
 export function parse(str, options = {}) {
   const tokeniser = new Tokeniser(str);
+  if (options.fileName) {
```
marcoscaceres:
```suggestion
  if (options.filename) {
```

lib/error.js
marcoscaceres: `name` could work :)

marcoscaceres: I'll delete my suggestions.

saschanaz:
>will we need to generate a name?

IMO yes, I'll probably just assign the index of `<pre>` array and use it to point the right elements.

marcoscaceres:
> IMO yes,

Sounds good. We can work that out on the ReSpec side.

saschanaz:
Work on it yesterday and failed to do xref thing there. Will open a PR without xref soon.

saschanaz:
Had to use `sourceName` instead because `name` in error.js conflicts with `Error.name`. Does it look good?

marcoscaceres:
sourceName is good too :)

saschanaz:
(Waiting on your approval)

PR #350, opened 2019-07-09
feat(validation): require [Exposed]

### LINKED ISSUE #342: Check existence of [Exposed] in validator
See also https://github.com/w3c/respec/issues/1370

saschanaz:
Closes #342


lib/productions/interface.js
```diff
@@ -40,4 +41,12 @@ export class Interface extends Container {
   get type() {
     return "interface";
   }
+
+  *validate(defs) {
+    if (!this.partial && this.extAttrs.every(extAttr => extAttr.name !== "Exposed")) {
+      const message = `Non-partial interfaces must have [Exposed] extended attribute.`;
```
marcoscaceres: Will this cover namespaces too?  Or do that separate?

saschanaz: This is only for interfaces, the namespace part is below in lib/productions/namespace.js 😀

marcoscaceres: Also, we want to  make these messages as friendly as possible, so we should include a "how to fix". So...  
```suggestion
      const message = `Interfaces and namespaces must have [Exposed] extended attribute. To fix, add, for example, [Exposed=(Window)]. Please also consider carefully if your interface should also be exposed in a Worker scope. Refer to the WebIDL spec section on Exposed for more information.`;
```

marcoscaceres: Ah, sorry, just saw your response about namespaces... didn't scroll down enough lol :)

lib/productions/namespace.js
```diff
@@ -24,4 +25,12 @@ export class Namespace extends Container {
   get type() {
     return "namespace";
   }
+
+  *validate(defs) {
+    if (!this.partial && this.extAttrs.every(extAttr => extAttr.name !== "Exposed")) {
+      const message = `Non-partial namespaces must have [Exposed] extended attribute.`;
```
marcoscaceres: As above :)

lib/productions/interface.js
marcoscaceres: We could even add direct links to the WebIDL spec using markdown.

PR #352, opened 2019-07-09
feat(validation): support batch validation

saschanaz:
A part of #347, so the batch validation can be more straightforward.

### DIFF
--- a/lib/validator.js
+++ b/lib/validator.js
@@ -102,7 +102,7 @@ function* validateIterable(ast) {
   yield* checkInterfaceMemberDuplication(defs);
 }
 
-// Remove this once all of our support targets exposes `.flat()` by default
+// Remove this once all of our support targets expose `.flat()` by default
 function flatten(array) {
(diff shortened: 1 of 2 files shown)


lib/validator.js
```diff
@@ -102,6 +102,16 @@ function* validateIterable(ast) {
   yield* checkInterfaceMemberDuplication(defs);
 }
 
+function flatten(array) {
```
saschanaz:
```suggestion
// Remove this once all of our support targets exposes `.flat()` by default
function flatten(array) {
```

lib/validator.js
```diff
@@ -102,6 +102,17 @@ function* validateIterable(ast) {
   yield* checkInterfaceMemberDuplication(defs);
 }
 
+// Remove this once all of our support targets exposes `.flat()` by default
```
saschanaz:
```suggestion
// Remove this once all of our support targets expose `.flat()` by default
```

PR #355, opened 2019-07-11
fix(lib/webidl2): respect falsy source names

saschanaz:


saschanaz:
Well, I should have written the write code... 😭

PR #365, opened 2019-07-15
feat: support async iterable<>

saschanaz:
And add `no-console` 😱

Implements https://github.com/heycam/webidl/pull/720


saschanaz:
Probably should split some of this if the actual merge takes some more time.

test/invalid/idl/async-iterable-single.webidl
```diff
@@ -0,0 +1,4 @@
+[Exposed=Window]
+interface AsyncIterable {
+  async iterable<long>;
```
marcoscaceres: Wonder if we should check:
async maplike<DOMString, DOMString>;

marcoscaceres:
Merged in WebIDL 🥳

PR #367, opened 2019-07-18
Add test case for default-initialized example

lukebjerring:
From https://heycam.github.io/webidl/#idl-operations

    The following IDL fragment defines an interface with an operation that takes a dictionary argument:
    
    dictionary LookupOptions {
      boolean caseSensitive = false;
    };
    
    [Exposed=Window]
    interface AddressBook {
      boolean hasAddressForName(USVString name, optional LookupOptions options = {});
    };
    If hasAddressForName is called with only one argument, the second argument will be a default-initialized LookupOptions dictionary, which will cause caseSensitive to be set to false.

Test case currently fails (expected to, as seems to be the cause of https://github.com/tidoust/reffy-reports/issues/34)

Not intending to land this PR without the fix, but it's a step in the right direction.

`default.json` content was generated by temporarily omitting the ` = {}` syntax, and JSON.stringify'ing the AST.


saschanaz:
Hi, thank you for the PR!

>default.json content was generated by temporarily omitting the = {} syntax, and JSON.stringify'ing the AST.

I think this is why the test is failing, could you explain the reason you did it?

And I'm not sure why we need another test as https://github.com/w3c/webidl2.js/pull/338 added one with `= {}`.

lukebjerring:
Well, for one, it's a different context; the test in #338 is a member with default value, not a default value of a parameter to an operator.

The AST fails to generate _at all_ with the `= {}` present, so I omitted it to generate at least something (to fill the json file). The failure mode for this test is not just a missing default, it's that the whole tree is missing.

Sorry if this reply is sloppy, I'm on my phone :)

lukebjerring:
In case it's not clear in the description:
See https://github.com/tidoust/reffy-reports/issues/34 for how this is manifesting problematically in a real-world application.

saschanaz:
>Well, for one, it's a different context; the test in #338 is a member with default value, not a default value of a parameter to an operator.

Ah, definitely worth then 👍 

>The AST fails to generate at all with the = {} present, so I omitted it to generate at least something (to fill the json file). The failure mode for this test is not just a missing default, it's that the whole tree is missing.

What did you do to generate AST? `npm run acquire` should autogenerate the JSON file, and I see it correctly generates an AST as expected. I'll push the result so that you can check it.

@marcoscaceres Do you have an idea why Travis is failing to start again? I thought the migration to service to app fixed it, but seems not 🤔

marcoscaceres:
Having a look at Travis...

marcoscaceres:
Hmm... I don't have rights to access the settings... but what's more strange is that the other PRs are all fine.

saschanaz:
I think Travis has a problem only with PRs from forks.

marcoscaceres:
Yeah, I was thinking the same... maybe just merge and hope for the best 🙊😂

lukebjerring:
Thanks!

PR #376, opened 2019-07-19
feat(validation): no dictionary in nullable union types

saschanaz:
Implements the latter part of [the rule](https://heycam.github.io/webidl/#idl-nullable-type):

>a union type that itself includes a nullable type or has a dictionary type as one of its flattened member types.


lib/productions/type.js
```diff
@@ -143,4 +145,52 @@ export class Type extends Base {
...
+        const message = `Nullable union cannot include a dictionary type`;
+        yield validationError(this.source, targetToken, this, message);
+      }
+    } else {
+      // allow some dictionary
+      for (const subtype of this.subtype) {
+        yield* subtype.validate(defs);
+      }
+    }
+    // if (this.nullable && this.union) {
```
marcoscaceres: keep the dead code?

saschanaz: Oops 💀

PR #377, opened 2019-07-23
Add file extensions

munchkinhalfling:
Brave Browser was giving me 404s and not loading the files at load because it didn't automatically add the JS file extension to imports, so I added `.js` to both of the imports.


saschanaz:
Thanks! I wonder there is a eslint rule to prevent this 🤔

Travis is currently broken so I'm manually meeting this.

PR #386, opened 2019-08-21
Setup GitHub Actions

saschanaz:
🎉

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -1,0 +1,24 @@
+name: Node CI
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [8.x, 10.x, 12.x]
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - name: npm install, build, and test
+      run: |
+        npm install
+        npm run build --if-present
+        npm test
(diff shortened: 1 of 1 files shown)


saschanaz:
The documentation is [here](https://help.github.com/en/articles/workflow-syntax-for-github-actions) but not sure how to mark the actions as "Required".

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,24 @@
...
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [8.x, 10.x, 12.x]
```
marcoscaceres: we should eventually just bind to LTS

saschanaz: Hmm, maybe this will follow the nvm input, I'll try it

saschanaz:
```suggestion
        node-version: [8.x, lts/*, node]
```

marcoscaceres: Any reason to keep testing on 8?

marcoscaceres: I think "lts" and "current" is fine. We shouldn't hold ourselves back unless we really need to.

saschanaz: It's just there because no need to remove it yet, as we haven't seen a failure there. I think we can freely remove it once we see one.

saschanaz:
Okay, the nvm syntax doesn't work. Did GitHub just invented their own node selector syntax? 🙄

saschanaz:
It's tracked by https://github.com/actions/setup-node/issues/26, will apply it when it's done.

marcoscaceres:
> It's tracked by actions/setup-node#26, will apply it when it's done.

Sounds good.

PR #388, opened 2019-08-21
feat(extAttrs): add autofix for Exposed=Window

saschanaz:
The second frequent invalid thing in IDL world.


saschanaz:
> as we do want authors to consider where they expose things. But I say we run with it :)

I think it's okay as autofixed results will still get reviewed by maintainers from each spec.

PR #391, opened 2019-08-24
chore(workflow): add pull_request to GitHub Actions

sidvishnoi:


saschanaz:
Heh, so it seems we need `pull_request`

PR #395, opened 2019-08-27
fix(productions/argument): no default for nullable dictionary args

saschanaz:
Currently the validator suggests adding a default dictionary value for a dictionary argument even it's nullable, but it's not a valid suggestion because a nullable dictionary type is not really a dictionary type. (https://github.com/heycam/webidl/issues/774#issuecomment-524229029)


saschanaz:
> Very cool! I wonder if we will see a lot of these?

Not sure I understand, but the autofix thing useful for my batch autofixer: https://github.com/saschanaz/webidl-updater

PR #400, opened 2019-08-30
fix(productions/operation): warn if no operation identifier

saschanaz:
[The syntax](https://heycam.github.io/webidl/#prod-OperationRest) somehow allows this, so this patch only warns when if no identifier [per the prose](https://heycam.github.io/webidl/#idl-operations):

>If an operation has no identifier, then it must be declared to be a special operation using one of the special keywords.

This patch includes:
- [x] A relevant test

### DIFF
--- a/lib/productions/operation.js
+++ b/lib/productions/operation.js
@@ -47,7 +47,7 @@ export class Operation extends Base {
 
   *validate(defs) {
     if (!this.name && ["", "static"].includes(this.special)) {
-      const message = `Regular or static operations must have an identifier.`;
+      const message = `Regular or static operations must have both a return type and an identifier.`;
       yield validationError(this.source, this.tokens.open, this, message);
     }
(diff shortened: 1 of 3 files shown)


lib/productions/operation.js
```diff
@@ -45,6 +46,10 @@ export class Operation extends Base {
   }
 
   *validate(defs) {
+    if (!this.name && ["", "static"].includes(this.special)) {
+      const message = `Regular or static operations need a proper identifier.`;
```
marcoscaceres: I think we should avoid "proper" as it's ambiguous. Maybe just: 

```suggestion
      const message = `Regular or static operations must have an identifier.`;
```

marcoscaceres: Maybe it should say "Did you forget the return type?"

saschanaz: That might be helpful.

PR #410, opened 2019-09-07
refactor(validator): reduce parameter count

saschanaz:
The first parameter was redundant.

### DIFF
--- a/lib/productions/interface.js
+++ b/lib/productions/interface.js
@@ -64,7 +64,7 @@ for more information.`;
     }
     const constructors = this.extAttrs.filter(extAttr => extAttr.name === "Constructor");
     for (const constructor of constructors) {
-      const message = `Constructors should now be represented as constructor operations \
+      const message = `Constructors should now be represented as a \`constructor()\` operation on the interface \
 instead of \`[Constructor]\` extended attribute. Refer to the \
 [WebIDL spec section on constructor operations](https://heycam.github.io/webidl/#idl-constructors) \
 for more information.`;
--- a/lib/productions/interface.js
+++ b/lib/productions/interface.js
@@ -58,17 +58,17 @@ To fix, add, for example, \`[Exposed=Window]\`. Please also consider carefully \
 if your interface should also be exposed in a Worker scope. Refer to the \
 [WebIDL spec section on Exposed](https://heycam.github.io/webidl/#Exposed) \
 for more information.`;
-      yield validationError(this.source, this.tokens.name, this, message, {
+      yield validationError(this.tokens.name, this, message, {
         autofix: autofixAddExposedWindow(this)
       });
     }
     const constructors = this.extAttrs.filter(extAttr => extAttr.name === "Constructor");
     for (const constructor of constructors) {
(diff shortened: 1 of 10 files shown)


lib/productions/interface.js
```diff
@@ -58,17 +58,17 @@ To fix, add, for example, \`[Exposed=Window]\`. Please also consider carefully \
...
 for more information.`;
-      yield validationError(this.source, this.tokens.name, this, message, {
+      yield validationError(this.tokens.name, this, message, {
         autofix: autofixAddExposedWindow(this)
       });
     }
     const constructors = this.extAttrs.filter(extAttr => extAttr.name === "Constructor");
     for (const constructor of constructors) {
-      const message = `Constructors should now represented as constructor operations \
+      const message = `Constructors should now be represented as constructor operations \
```
marcoscaceres:
```suggestion
      const message = `Constructors should now be represented as a `constructor()` operation on the interface \
```

lib/productions/interface.js
```diff
@@ -58,17 +58,17 @@ To fix, add, for example, \`[Exposed=Window]\`. Please also consider carefully \
...
 instead of \`[Constructor]\`. Refer to the \
```
marcoscaceres:
```suggestion
instead of \`[Constructor]\` extended attribute. Refer to the \
```

lib/productions/interface.js
saschanaz: `constructor()` constructor operation? 🤔

marcoscaceres: heh, oops.. let me fix :D

marcoscaceres: Actually, it's ok...

"Constructors should now be represented as a `constructor()` operation on the interface instead of `[Constructor]` extended attribute.

Alternative:
The `[Constructor]` extended attribute is deprecated. Please use a `constructor()` operation directly on the interface instead.

marcoscaceres: And if we have the name of the interface that we can dynamically include that would be super nice.

marcoscaceres: "directly on the `Foo` interface instead" 😍

saschanaz: It's not exactly "deprecated" though, it's "removed" 😁

> And if we have the name of the interface that we can dynamically include that would be super nice.

It's included in the full validation message: https://github.com/w3c/webidl2.js/pull/410/files#diff-db228a9a69bb3fb15e5a4e1ed87a9080L1

saschanaz: My small concern is that `constructor()` is not a general operation but a "constructor operation" and it's type is "constructor" instead of "operation".

marcoscaceres: heh, true... I was being polite... but being direct is and saying it's been removed is better.

marcoscaceres: > My small concern is that constructor() is not a general operation but a "constructor operation" and it's type is "constructor" instead of "operation".
 
It's a tough one, because very few people actually read the WebIDL spec so the terminology might confuse them. We want to provide the minimum amount of information to say "there is a problem. Here is how you fix it quickly: "

PR #419, opened 2019-09-23
fix(productions/interface): correct constructor indentation

saschanaz:
There was a bug that `constructor()` get too much indentation or too less.

This patch includes:
- [x] A relevant test

### DIFF
--- a/test/autofix.js
+++ b/test/autofix.js
@@ -112,6 +112,7 @@ describe("Writer template functions", () => {
        Exposed=Window]
       interface B {
         attribute any attr;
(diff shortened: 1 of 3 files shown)


test/autofix.js
```diff
@@ -112,6 +112,7 @@ describe("Writer template functions", () => {
        Exposed=Window]
       interface B {
         attribute any attr;
+        // attribute any popipa;
```
marcoscaceres: heh :)

saschanaz: 🤣

PR #420, opened 2019-09-23
fix(productions/helpers): emit newline after [Exposed]

saschanaz:
More frequent formatting.

PS: The `getLastIndentation` part is copy-pasted from #419.

This patch includes:
- [x] A relevant test

### DIFF
--- a/lib/productions/helpers.js
+++ b/lib/productions/helpers.js
@@ -170,10 +170,11 @@ export function stringifier(tokeniser) {
 /**
  * @param {string} str
  */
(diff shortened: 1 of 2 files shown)


saschanaz:
> I can't recall, but can Expose also appear on attributes? If yes, worth adding a test for that?

Yes, it can appear. No, no test needed, because it's completely optional so we don't automatically add one for attributes.

saschanaz:
Maybe it should be `fix(autofix):` instead to better understand the purpose.

PR #423, opened 2019-09-24
fix(autofix): properly detect operation indentation

saschanaz:
Because operations does not have keywords in front of them.

This patch includes:
- [x] A relevant test


saschanaz:
> heh, eventually we could have an option to fix indenting (replace tabs with spaces, etc.) like in ESLint.

Probably! Not on my radar right now, though.

PR #425, opened 2019-09-28
feat: add `.parent` field

saschanaz:
WIP because I want everything to be less complex than this.

This patch includes:
- [x] A relevant test

### DIFF
--- a/README.md
+++ b/README.md
@@ -535,7 +535,7 @@ The fields are as follows:
 * `name`: The name of the operation if exists.
 * `arguments`: An array of [arguments](#arguments) for the operation.
 * `extAttrs`: An array of [extended attributes](#extended-attributes).
-* `parent`: The container of this type.
+* `parent`: The container of this type as an Object.
 
 ### Constructor Operation Member
 
@@ -568,7 +568,7 @@ The fields are as follows:
 * `type`: Always "constructor".
 * `arguments`: An array of [arguments](#arguments) for the constructor operation.
 * `extAttrs`: An array of [extended attributes](#extended-attributes).
-* `parent`: The container of this type.
+* `parent`: The container of this type as an Object.
 
 ### Attribute Member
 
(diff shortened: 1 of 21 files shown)


saschanaz:
I think the patch is now simple enough, but somewhat magical... What do you think, @marcoscaceres ?

README.md
```diff
@@ -550,6 +535,7 @@ The fields are as follows:
 * `name`: The name of the operation if exists.
 * `arguments`: An array of [arguments](#arguments) for the operation.
 * `extAttrs`: An array of [extended attributes](#extended-attributes).
+* `parent`: The container of this type.
```
marcoscaceres: Maybe: 

```suggestion
* `parent`: The container of this type as an Object.
```

Or anything else that might aid in the usage. I guess they are of whatever parent type. I guess this comment applies to the rest of the documentation.

lib/productions/helpers.js
```diff
@@ -242,3 +242,43 @@ export function findLastIndex(array, predicate) {
...
+
+/**
+ * Returns a proxy that auto-assign `parent` field.
+ * @template T
+ * @param {T} data
+ * @param {*} parent this is `data` by default
+ * @return {T}
+ */
+export function autoParenter(data, parent) {
+  if (!parent) {
```
marcoscaceres: Might be nice to document what's happening here (and below).... this is really nice,  but you are right that it's a bit magical. I'd be a bit scared to come in here in the future if anything goes wrong.

lib/productions/helpers.js
```diff
@@ -242,3 +242,43 @@ export function findLastIndex(array, predicate) {
...
+    parent = data;
+  }
+  if (!data || typeof data !== "object") {
```
marcoscaceres: Similarly here... explaining why this is not "exceptional" would be good (I guess this is parent root or something).

lib/productions/helpers.js
```diff
@@ -242,3 +242,43 @@ export function findLastIndex(array, predicate) {
...
+    return data;
+  }
+  return new Proxy(data, {
+    get(target, p) {
+      const value = target[p];
+      if (Array.isArray(value)) {
+        return autoParenter(value, target);
```
marcoscaceres: Explaining this one a little bit would be good too... recursive array fun :)

Issue #220, opened 2018-06-25
Export error constructor


saschanaz:
... so that users can do `err instanceof WebIDLParseError`.

Alexkho82:
#220

saschanaz:
Closed by #428

PR #428, opened 2019-09-29
Solves issue 220

TG1999:
#220 solved
This patch includes:
- [x] A relevant test


lib/tokeniser.js
```diff
@@ -251,3 +251,6 @@ class WebIDLParseError extends Error {
     this.tokens = tokens;
   }
 }
+module.exports={
```
saschanaz: This is old CommonJS way to export things, here instead we want to use more modern way: just add `export` before `class WebIDLParseError` 👍

index.js
```diff
@@ -1,3 +1,4 @@
 export { parse } from "./lib/webidl2.js";
 export { write } from "./lib/writer.js";
 export { validate } from "./lib/validator.js";
+export {Err} from "./lib/tokeniser.js";
```
saschanaz: This is good, but I would prefer the original name `WebIDLParseError`.

saschanaz: Make sure we get spaces between braces!

index.js
```diff
@@ -1,3 +1,4 @@
 export { parse } from "./lib/webidl2.js";
 export { write } from "./lib/writer.js";
 export { validate } from "./lib/validator.js";
+export {WebIDLParseError} from "./lib/tokeniser.js";
```
saschanaz:
```suggestion
export { WebIDLParseError } from "./lib/tokeniser.js";
```

For code format consistency 😁 (You can click "Commit suggestion" to automatically accept my suggestion)

lib/tokeniser.js
```diff
@@ -251,3 +251,4 @@ class WebIDLParseError extends Error {
     this.tokens = tokens;
   }
 }
+
```
saschanaz:
```suggestion
```

Please remove redundant newline here!

index.js
TG1999: Okay I am adding spaces now it is also passing all the tests, so I think it can be merged after that change :)

TG1999:
@saschanaz all the changes requested in review are done so I think you can merge it now :)

saschanaz:
Code now looks great, but we have slightly more longer work here: Adding a test!

We want to add a test in [test/invalid.js](https://github.com/w3c/webidl2.js/blob/gh-pages/test/invalid.js), where we want to make sure `err instanceof WebIDLParseError` (which is our original purpose) works as intended.

At line 131, we want to add a new line `it("allows `instanceof WebIDLParseError`", () => {` and check it works as described. `parse("throwerror")` will throw the error we exported just now, so you'll want to try-catch that error and `expect(err).toBeInstanceOf(WebIDLParseError)`.

TG1999:
Okay on it

TG1999:
@saschanaz any suggestions why this is failing ? Somngthing regarding whitespace error is given at 132:1 I am not able to understand it

saschanaz:
I don't see a trailing whitespace there... but try `npm run lint -- --fix` and it will automatically try fixing the error, if you have a local clone of this repo.

test/invalid.js
```diff
@@ -129,6 +129,16 @@ describe("Error object structure", () => {
     const { level } = validation[0];
     expect(level).toBe("warning");
   });
+
```
saschanaz:
```suggestion

```

Oh, I got it, we do have trailing spaces here.

TG1999:
Okay if I will fix at my local too, it will still give an error here, so how will we pass the tests here ?

test/invalid.js
```diff
@@ -129,6 +129,16 @@ describe("Error object structure", () => {
...
+  
+  it("allows `instanceof WebIDLParseError`", () => {
+    try {
+      parse("throwerror");
+      throw new Error("Shouldn't reach here");
+    } catch ({err}) {
+      expect(err).toBeInstanceOf(WebIDLParseError);
+    }
+  });
+
```
saschanaz:
```suggestion
```

This line looks redundant.

TG1999:
Sorry I closed it by mistake, just making the changes. :)

TG1999:
Hey @saschanaz please can you have  look again :sweat_smile:

test/invalid.js
```diff
@@ -129,6 +129,14 @@ describe("Error object structure", () => {
     const { level } = validation[0];
     expect(level).toBe("warning");
   });
+  it("allows `instanceof WebIDLParseError`", () => {
```
saschanaz:
```suggestion

  it("allows `instanceof WebIDLParseError`", () => {
```

Oops, we accidentally removed the newline before this line 😲

test/invalid.js
```diff
@@ -129,6 +129,14 @@ describe("Error object structure", () => {
...
+    try {
+      parse("throwerror");
+      throw new Error("Shouldn't reach here");
+    } catch ({err}) {
```
saschanaz: `{err}` accesses `err` field of the error object, I think we just want `err` to get the error object itself.

TG1999: Okay I have made the changes and they have passed. Thanks for your help :smile:

Issue #430, opened 2019-09-30
Disallow Promise types inside unions


saschanaz:
See also https://github.com/heycam/webidl/pull/746

saschanaz:
@TG1999 Wanna try this one?

We currently disallows `any` inside a union, but now we also want to disallow `Promise<>` there. The relevant line is:

https://github.com/w3c/webidl2.js/blob/7fb1c2f1c4fb5390e9e0239cb3978bbe4f30cc4c/lib/productions/type.js#L93

We probably want to check `if (typ.generic === "Promise")` and then throw.

The test for this is different from what you did yesterday. We want to:

* Add a file in [test/invalid/idl/](https://github.com/w3c/webidl2.js/tree/gh-pages/test/invalid/idl) that includes a wrong syntax. `typedef (Promise<DOMString> or boolean) Typedef;` should be sufficient.
* Run `npm run acquire` to generate baseline files
* Check that we got a new file in `test/invalid/baseline` with the expected error message.

TG1999:
Yeah sure I don't have access to my system right now, will start work on it after 12 hours.

TG1999:
Hi I am back ,I am currently adding line 94 as if (typ.generic==="Promise") tokeniser.error("Type `any` cannot be included in a union type");

saschanaz:
Closed by #435

PR #431, opened 2019-09-30
promise disallowed

TG1999:

This patch includes:
- [ ] A relevant test


TG1999:
@saschanaz please have a look on this

TG1999:
If it looks good then we can start test part in #430

lib/productions/type.js
```diff
@@ -91,6 +91,7 @@ function union_type(tokeniser, type) {
   while (true) {
     const typ = type_with_extended_attributes(tokeniser) || tokeniser.error("No type after open parenthesis or 'or' in union type");
     if (typ.idlType === "any") tokeniser.error("Type `any` cannot be included in a union type");
+    if (typ.generic === "Promise") tokeniser.error("Type `Promise` cannot be included in a union type");
```
saschanaz: This is already good!

But could we move this two lines (line 93 and 94) before line 92, and change them into `if (tokeniser.probe("any"))` and `if (tokeniser.probe("Promise"))` instead of current one? That would create a better error message.

test/invalid/baseline/typedef-promise.txt
```diff
@@ -0,0 +1 @@
+
```
saschanaz: Could you `npm run acquire` again and commit the result?

test/invalid/idl/typedef-promise.webidl
```diff
@@ -0,0 +1 @@
+typedef (Promise<DOMString> or boolean) Typedef;
```
saschanaz: Could you change the filename into `union-promise` because that's the purpose of this patch?

TG1999:
I have ran npm run acquire before it generated empty file, that's why it is giving error

saschanaz:
> I have ran npm run acquire before it generated empty file, that's why it is giving error

Oops sorry, the correct step is `npm run build && npm run acquire` because we need the build result to run.

TG1999:
please have a look at PR #435 I was facing some issue in git that's why created a new branch

PR #432, opened 2019-10-01
chore(package): update deps

saschanaz:


marcoscaceres:
I guess we never set up greenkeeper for this project? should we do that?

saschanaz:
That would be helpful.

Issue #85, opened 2017-10-17
parent of things


marcoscaceres:
When processing IDL, it would be nice have a back reference to the parent/owner of things from the parse tree. 

Like, "operation" would have the parent "interface" as a reference. Enum values would have the Enum as the parent. Params would have the "operation" as the parent... and so on up the tree.

Interfaces, Enums, Dictionaries would just be `null` as the parent.

marcoscaceres:
To be clear, I want to be able to do:

```
const interfaceObj = operation.parent;
if(interfaceObj.name === "Whatever"){...}; 
```

SaschaNaz:
This will help implementing "[The [Replaceable] extended attribute must not be used on an attribute declared on a callback interface](https://heycam.github.io/webidl/#Replaceable)" restriction 👍

marcoscaceres:
Played a bit with how to do this... but we might not be able to do this without significant changes (to tests): this would create a cyclic structure, breaking JSON output.

SaschaNaz:
YAML is capable of cycle structure and jsondiffpatch also can cope with that. https://runkit.com/embed/mjk8jwzlgydl

I think transitioning to YAML is a viable choice, what do you think?

marcoscaceres:
I'm ok to live without the parents (and happy to live with JSON for now). I wrote this algo to get around the limitation:

```JS
function addParent(obj, owner = null) {
  obj.parent = owner;
  Array.from(Object.entries(obj))
    .filter(([key]) => key !== "parent")
    .filter(([, value]) => Array.isArray(value) || value instanceof Object)
    .map(([, value]) => value)
    .forEach(value => {
      if (Array.isArray(value)) {
        value
          .filter(({ type }) => type !== "," && !whiteSpaceTypes.has(type))
          .forEach(item => {
            addParent(item, obj);
          });
        return;
      }
      value.parent = obj;
    });
}
```

marcoscaceres:
Closing as won’t fix for now.

marcoscaceres:
Continue to find this really useful in ReSpec and in the validator - having the parent gives really context ("The foo attribute of Interface Bar is missing X" and so one)... what I might do is exclude the parent when `JSON.stringify()` is called with the tests.

SaschaNaz:
We can't test it then, an untested feature is prone to break. I think we ultimately should be able to test cyclic structures.

SaschaNaz:
[...]

PR #434, opened 2019-10-02
feat(validator): check dictionary argument optionality

### LINKED ISSUE #407: Validate optionality of dictionary arguments
This is intended to catch issues like the one in https://github.com/w3c/performance-timeline/issues/119, fixed by https://github.com/w3c/performance-timeline/pull/120.

Here's a snippet of code that should cause validation errors:
```js
const { parse, validate } = require("webidl2");
const tree = parse(`
dictionary BrilliantInit {
  any stuff;
};

[Exposed=Window]
interface X {
    constructor(BrilliantInit dict);
};
`);
const validation = validate(tree);
console.log(validation);
```

The `dict` argument there should be optional.

It might also be a good idea to ensure that if a dictionary has required members, then it's *not* optional. However, https://heycam.github.io/webidl/#idl-operations doesn't seem to require this...

@bzbarsky

saschanaz:
Closes #407 

WIP, probably depends on #85 to check argument indices.

This patch includes:
- [x] A relevant test

### DIFF
--- a/test/autofix.js
+++ b/test/autofix.js
@@ -262,4 +262,32 @@ describe("Writer template functions", () => {
     `;
     expect(autofix(inputMultiple)).toBe(outputMultiple);
   });
+
+  it("should add `optional` for non-required dictionary arguments", () => {
+    const input = `
+      dictionary Optional {
+        long long loooongInterviewProcess;
+      };
+
+      interface mixin Container {
+        void op(
+          DOMString str,
+          Optional arg
+        );
+      };
+    `;
+    const output = `
+      dictionary Optional {
+        long long loooongInterviewProcess;
+      };
+
+      interface mixin Container {
+        void op(
+          DOMString str,
+          optional Optional arg = {}
+        );
+      };
+    `;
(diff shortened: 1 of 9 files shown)


test/util/collect.js
```diff
@@ -28,7 +28,7 @@ function* collect(base, { expectError, raw } = {}) {
       }
     }
     catch (error) {
-      if (expectError) {
+      if (expectError && error instanceof wp.WebIDLParseError) {
```
saschanaz: Thanks you @TG1999 for #428  🎉🎉

test/autofix.js
```diff
@@ -262,4 +262,32 @@ describe("Writer template functions", () => {
     `;
     expect(autofix(inputMultiple)).toBe(outputMultiple);
   });
+
+  it("should add `optional` for non-required dictionary arguments", () => {
+    const input = `
+      dictionary Optional {
+        long long loooongInterviewProcess;
```
marcoscaceres: 😂💀🔥

foolip: Classic :)

PR #437, opened 2019-10-02
fix(autofix): correct constructor ordering 2

saschanaz:
It was broken last time 🤔

This patch includes:
- [x] A relevant test


saschanaz:
GitHub is weird today, waiting a ghost...

Issue #256, opened 2019-01-14
parse() throws error on valid ReflectOnly Blink extended attributes


jpmedley:
According to the Blink IDL Extended Attributes spec, [ReflectOnly has two valid forms](https://chromium.googlesource.com/chromium/src/+/master/third_party/blink/renderer/bindings/IDLExtendedAttributes.md#ReflectOnly_a):

    [Reflect, ReflectOnly="on"] attribute DOMString toggle;
    // Parens may contain 2 to n values.
    [Reflect=q, ReflectOnly=("anonymous", "use-credentials")] attribute DOMString quarter;

The second format is causing error such as the following to be thrown:

    Got an error during or right after parsing `interface HTMLImageElement`: Expected
    identifiers but none found, line 33 (tokens: "\"anonymous\",\"use-credentials\"),")

There are 11 files in Chromium (listed below) that trigger this error. I've [already verified](https://bugs.chromium.org/p/chromium/issues/detail?id=920009) that the Blink IDL spec is correct.

blink/renderer/core/html/html_element.idl
blink/renderer/core/html/html_link_element.idl
blink/renderer/core/html/html_table_cell_element.idl
blink/renderer/core/html/html_anchor_element.idl
blink/renderer/core/html/forms/html_form_element.idl
blink/renderer/core/html/html_area_element.idl
blink/renderer/core/html/html_script_element.idl
blink/renderer/core/html/html_iframe_element.idl
blink/renderer/core/html/html_image_element.idl
blink/renderer/core/html/media/html_media_element.idl
blink/renderer/core/dom/element.idl

I would appreciate some haste in addressing this. This affects tooling I've built to increase my velocity at [documenting new web platform APIs](https://developer.mozilla.org/en-US/profiles/jpmedley). If I can assist in some further way, will be happy to do so.

Thanks

jyasskin:
The ReflectOnly Blink extended attribute has a form different from the [five forms used in the WebIDL spec](https://heycam.github.io/webidl/#idl-extended-attributes). It's closest to _ExtendedAttributeIdentList_, but it takes strings instead of tokens. https://github.com/w3c/webidl2.js/blob/develop@{01-14-19}/lib/webidl2.js#L595 says webidl2.js takes the shortcut of parsing less than the spec actually allows, so it should probably grow new support for this sixth form.

saschanaz:
I think it's worth supporting the string extended-form as it seems it's being used frequently enough. Thoughts? @marcoscaceres 

[...]

Issue #251, opened 2018-12-09
BufferRelatedType shouldn't be read as identifiers


saschanaz:
https://heycam.github.io/webidl/#index-prod-BufferRelatedType

```
BufferRelatedType ::
    ArrayBuffer
    DataView
    Int8Array
    Int16Array
    Int32Array
    Uint8Array
    Uint16Array
    Uint32Array
    Uint8ClampedArray
    Float32Array
    Float64Array
```

These are keywords so the following is illegal:

```
interface Int32Array {};
```

saschanaz:
BufferRelatedType is not exclusively used anywhere, though, whereas StringType is.

saschanaz:
Also closed by #439

PR #439, opened 2019-10-03
fix: `any` not a keyword

### LINKED ISSUE #436: `any` is not a keyword
But it should be!

TG1999:

This patch closes #436  and includes:
- [ ] A relevant test


lib/tokeniser.js
```diff
@@ -72,7 +72,8 @@ const nonRegexTerminals = [
   "short",
   "true",
   "unsigned",
-  "void"
+  "void",
+  "any"
```
saschanaz: I would keep this in alphabet order, so probably line 56.

TG1999:
Hi @saschanaz I have wrote one test case for any and one for int32 I am opening another issue for writing test cases for solving #251 . Please merge this one

lib/tokeniser.js
```diff
@@ -53,6 +69,7 @@ const nonRegexTerminals = [
   "Infinity",
   "NaN",
   "Promise",
+  "any",
   "async",
```
saschanaz: Again while you are at it, could you remove `async` and `constructor` from this list? They are already in `argumentNameKeywords` so they are redundant.

lib/tokeniser.js
```diff
@@ -15,6 +15,22 @@ const tokenRe = {
   "other": /[^\t\n\r 0-9A-Za-z]/y
 };
 
+export const typeNameKeywords=[
```
saschanaz:
```suggestion
export const typeNameKeywords = [
```

Make sure the code format be consistent with other lines 😁

lib/productions/type.js
```diff
@@ -62,7 +62,7 @@ function type_suffix(tokeniser, obj) {
 function single_type(tokeniser, typeName) {
   let ret = generic_type(tokeniser, typeName) || primitive_type(tokeniser);
   if (!ret) {
-    const base = tokeniser.consume("identifier", ...stringTypes);
+    const base = tokeniser.consume("identifier", "any", ...stringTypes, ...typeNameKeywords);
```
saschanaz: This is okay, but we could probably move `any` into `typeNameKeywords` 👍

test/invalid/baseline/Int32-keyword.txt
```diff
@@ -0,0 +1,3 @@
+Syntax error at line 1 in Int32-keyword.webidl:
```
saschanaz: 👍 but could you rename this as `int32array-keyword` because

* I would keep filenames in lowercase, just for consistency
* We may get `int32` in the future, so better be more specific

Sorry to be picky here 😱

TG1999:
@saschanaz all changes made :grinning:

saschanaz:
If you want to continue your great works, #256 seems to be next easiest issue to fix.

It might be more complex than previous issues, we want to modify parser logic to allow a set of string in extended attributes. (In `lib/productions/extended-attribute.js`)

Probably we'll want to copy `identifiers()` function there and make it parse strings.

I can provide more details but it's already time to bed, so if you need more information, you might try reading existing code before I return.

TG1999:
Hi @saschanaz I will surely like to work on this

TG1999:
I have read the existing code from their I understood ```export function identifiers(tokeniser) {
  const ids = list(tokeniser, { parser: Token.parser(tokeniser, "identifier"), listName: "identifier list" });
  if (!ids.length) {
    tokeniser.error("Expected identifiers but none found");
  }
  return ids;
}``` I have to implement this functionality In lib/productions/extended-attribute.js so from where I will get the string that I have to parse

saschanaz:
I would modify `identifiers` into `tokens` and add a new function `identifiersOrStrings`, all in `extended-attribute.js`:

```js
function tokens(tokeniser, tokenName) {
  const toks = list(tokeniser, {
    parser: Token.parser(tokeniser, tokenName),
    listName: tokenName + " list"
  });
  if (!toks.length) {
    tokeniser.error(`Expected ${tokenName}s but none found`);
  }
  return toks;
}

function identifiersOrStrings(tokeniser) {
  return tokens(tokeniser, "identifier") || tokens(tokeniser, "string");
}
```

And replace existing `identifiers(tokeniser)` into `identifiersOrStrings(tokeniser)`. This will allow `("str1", "str2")` where we want.

TG1999:
Okay I got it @saschanaz , will be making a PR soon then we can discuss more changes on the PR itself.

saschanaz:
Looking forward to the PR! (If you are blocked on anything, I'm always here to help 👍)

TG1999:
Hi, right now I am stucked with some college work, gill try to get back to you by weekend and thanks for your help 😄

PR #435, opened 2019-10-02
promises disallowed

TG1999:
Signed-off-by: tushar goel <tushar.goel.dav@gmail.com>

This patch includes:
- [ ] A relevant test
- [ ] A relevant documentation update

PR #445, opened 2019-10-13
fix parse() throws error on valid ReflectOnly Blink extended attributes

### LINKED ISSUE #256: parse() throws error on valid ReflectOnly Blink extended attributes

TG1999:

This patch closes #256  and includes:
- [x] A relevant test

### DIFF
--- a/lib/productions/extended-attributes.js
+++ b/lib/productions/extended-attributes.js
@@ -13,15 +13,20 @@ function tokens(tokeniser, tokenName) {
     parser: Token.parser(tokeniser, tokenName),
     listName: tokenName + " list"
   });
-  if (!toks.length) {
-    tokeniser.error(`Expected ${tokenName}s but none found`);
-  }
   return toks;
 }
 
 // This will allow a set of strings to be parsed
  function identifiersOrStrings(tokeniser) {
-  return tokens(tokeniser, "identifier") || tokens(tokeniser, "string");
+  let toks = tokens(tokeniser, "identifier");
+  if (toks.length) {
+    return toks;
+  }
+  toks = tokens(tokeniser, "string");
+  if (toks.length) {
+    return toks;
+  }
+  tokeniser.error(`Expected identifiers or strings but none found`);
 }
 
 
--- a/lib/productions/extended-attributes.js
+++ b/lib/productions/extended-attributes.js
@@ -18,6 +18,7 @@ function tokens(tokeniser, tokenName) {
 
 /**
  * This will allow a set of identifiers or strings to be parsed.
+ * @param {import("../tokeniser").Tokeniser} tokeniser
  */
 function identifiersOrStrings(tokeniser) {
   let toks = tokens(tokeniser, "identifier");
--- a/lib/productions/extended-attributes.js
+++ b/lib/productions/extended-attributes.js
@@ -1,7 +1,29 @@
 import { Base } from "./base.js";
 import { ArrayBase } from "./array-base.js";
-import { list, identifiers, argument_list, autoParenter } from "./helpers.js";
+import { Token } from "./token.js";
+import { list, argument_list, autoParenter } from "./helpers.js";
 import { validationError } from "../error.js";
+
+/**
+ * @param {import("../tokeniser").Tokeniser} tokeniser
+ * @param {string} tokenName
+ */
+function tokens(tokeniser, tokenName) {
+  const toks = list(tokeniser, {
+    parser: Token.parser(tokeniser, tokenName),
+    listName: tokenName + " list"
+  });
+  if (!toks.length) {
+    tokeniser.error(`Expected ${tokenName}s but none found`);
+  }
+  return toks;
+}
+
+// This will allow a set of strings to be parsed
+ function identifiersOrStrings(tokeniser) {
+  return tokens(tokeniser, "identifier") || tokens(tokeniser, "string");
+}
+
 
 class ExtendedAttributeParameters extends Base {
   /**
@@ -17,7 +39,7 @@ class ExtendedAttributeParameters extends Base {
     if (tokens.open) {
       ret.list = ret.rhsType === "identifier-list" ?
         // [Exposed=(Window,Worker)]
-        identifiers(tokeniser) :
+        identifiersOrStrings(tokeniser) :
         // [NamedConstructor=Audio(DOMString src)] or [Constructor(DOMString str)]
         argument_list(tokeniser);
       tokens.close = tokeniser.consume(")") || tokeniser.error("Unexpected token in extended attribute argument list");
--- a/lib/productions/extended-attributes.js
+++ b/lib/productions/extended-attributes.js
@@ -16,7 +16,9 @@ function tokens(tokeniser, tokenName) {
   return toks;
 }
 
-// This will allow a set of strings to be parsed
+/**
+ * This will allow a set of identifiers or strings to be parsed.
+ */
 function identifiersOrStrings(tokeniser) {
   let toks = tokens(tokeniser, "identifier");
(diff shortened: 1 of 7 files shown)


TG1999:
@saschanaz looks good to merge ?

saschanaz:
Yes, given that we have a baseline test 👍

-Kagami

________________________________

TG1999:
@saschanaz can you guide me a little how can I start proceed to write a baseline test for this ?

saschanaz:
Sure! For now just copy-pasting the code in the OP of #256 should be sufficient.

-Kagami

________________________________

TG1999:
Can you highlight the required comment for me, so we can close this issue ASAP and start working on another issue.

saschanaz:
Sorry, I should have provided more detail. The code would be:

```webidl
interface mixin Reflector {
  [Reflect, ReflectOnly="on"] attribute DOMString toggle;
  // Parens may contain 2 to n values.
  [Reflect=q, ReflectOnly=("anonymous", "use-credentials")] attribute DOMString quarter;
};
```

Recall what you did in #435 👍

TG1999:
@saschanaz Done with the tests.

lib/productions/extended-attributes.js
```diff
@@ -1,8 +1,30 @@
...
+ * @param {import("../tokeniser").Tokeniser} tokeniser
+ * @param {string} tokenName
+ */
+function tokens(tokeniser, tokenName) {
+  const toks = list(tokeniser, {
+    parser: Token.parser(tokeniser, tokenName),
+    listName: tokenName + " list"
+  });
+  if (!toks.length) {
+    tokeniser.error(`Expected ${tokenName}s but none found`);
```
saschanaz: Could you move this length check to `identifiersOrStrings()? Otherwise it won't consume strings as it fails consuming identifiers first.

TG1999: Since we are getting toks after parsing them, so should I also parse it first?

saschanaz: Exactly.

TG1999:
@saschanaz I am moving it to syntax then I am getting this error while npm run acquire
/webidl2.js/test/util/collect.js:35
        throw error;
        ^

WebIDLParseError: Syntax error at line 4 in reflector-interface.webidl, since `interface mixin Reflector`:
q, ReflectOnly=("anonymous", "use-credentials")]
                ^ Expected identifiers but none found

saschanaz:
Maybe you have to `npm run build && npm run acquire`?

lib/productions/extended-attributes.js
TG1999: Then I think token functions is of no use

TG1999:
> Maybe you have to `npm run build && npm run acquire`?

@saschanaz still getting same error

saschanaz:
Oh, I incorrectly assumed `tokens()` would return a falsy value. `||` won't work, instead we should probably do something like:

```jjs
function identifiersOrStrings(tokeniser) {
  let toks = tokens(tokeniser, "identifier");
  if (toks.length) {
    return toks;
  }
  toks = tokens(tokeniser, "string");
  if (toks.length) {
    return toks;
  }
  tokeniser.error(`Expected identifiers or strings but none found`);
}
```

TG1999:
@saschanaz I think we should have it in invalid not in syntax. what say ?

saschanaz:
The new error messages suggest all is well, you can put it in `syntax` and you won't see any failure this time.

TG1999:
@saschanaz please have a look

saschanaz:
Hmm, please commit `npm run build && npm run acquire` (without `dist/`) and those will be fixed.

lib/productions/extended-attributes.js
```diff
@@ -1,8 +1,35 @@
...
+function tokens(tokeniser, tokenName) {
+  const toks = list(tokeniser, {
+    parser: Token.parser(tokeniser, tokenName),
+    listName: tokenName + " list"
+  });
+  return toks;
+}
+
+// This will allow a set of strings to be parsed
+ function identifiersOrStrings(tokeniser) {
```
saschanaz:
```suggestion
function identifiersOrStrings(tokeniser) {
```

One nit 😁

lib/productions/extended-attributes.js
```diff
@@ -1,8 +1,35 @@
...
+ */
+function tokens(tokeniser, tokenName) {
+  const toks = list(tokeniser, {
+    parser: Token.parser(tokeniser, tokenName),
+    listName: tokenName + " list"
+  });
+  return toks;
+}
+
+// This will allow a set of strings to be parsed
```
saschanaz:
```suggestion
/**
 * This will allow a set of identifiers and strings to be parsed.
 */
```

This way IDEs can provide a description when a dev hovers on a relevant function call.

TG1999:
> Hmm, please commit `npm run build && npm run acquire` (without `dist/`) and those will be fixed.

Doing that is giving me error which I have talked about earlier.

saschanaz:
> Doing that is giving me error which I have talked about earlier.

I tried it myself and it went as expected, did you retried after your last commit?

TG1999:
Yes

saschanaz:
Okay, given that you already wrote good codes, let me commit the result...

TG1999:
@saschanaz looks good to merge now ?

lib/productions/extended-attributes.js
```diff
@@ -1,8 +1,37 @@
...
+function tokens(tokeniser, tokenName) {
+  const toks = list(tokeniser, {
+    parser: Token.parser(tokeniser, tokenName),
+    listName: tokenName + " list"
+  });
+  return toks;
+}
+
+/**
+ * This will allow a set of identifiers or strings to be parsed.
```
saschanaz: Last nit:

```suggestion
 * This will allow a set of identifiers or strings to be parsed.
 * @param {import("../tokeniser").Tokeniser} tokeniser
```

TG1999:
No problem @saschanaz and thanks for guiding me 😊

PR #446, opened 2019-10-13
fix(productions/extended-attributes): introduce string-list

saschanaz:
Because strings in identifier-list are confusing. (Followup after #445)

This patch includes:
- [x] A relevant test


saschanaz:
@marcoscaceres Could you review this too so that we can solve this before a new release? 🙏

PR #456, opened 2020-02-03
feat(extended-attributes): support decimals/integers

### LINKED ISSUE #455: More permissive extended attributes grammar
We would like to use some extended attribute types in jsdom, but that webidl2.js cannot currently parse.

```webidl
interface HTMLMyElement {
  // List of strings (from https://chromium.googlesource.com/chromium/src/+/master/third_party/blink/renderer/bindings/IDLExtendedAttributes.md#reflectonly_a)
  [Reflect, ReflectOnly=("first", "second", "third", "fourth")] attribute DOMString quarter;

  // List of integers
  [Reflect, ReflectRange=(1, 4)] attribute unsigned long span;
};
```

----

#222 looked into changing the extended attributes parser to be more restrictive. heycam/webidl#574 resolved the question by saying that the grammar is intentionally generic. Indeed, the current [_ExtendedAttribute_](https://heycam.github.io/webidl/#prod-ExtendedAttribute) grammar allows for basically anything given that (quoting from #222):

> Any types of brackets must be closed if appears. Commas are allowed only inside brackets (as any comma outside will be a separator for the extended attribute).

saschanaz:
This better conforms with the permissiveness of extended attributes.

This patch closes #455 and includes:
- [x] A relevant test

### DIFF
--- a/test/syntax/idl/extended-attributes.webidl
+++ b/test/syntax/idl/extended-attributes.webidl
@@ -7,7 +7,7 @@ interface ServiceWorkerGlobalScope : WorkerGlobalScope {
 
 // Conformance with ExtendedAttributeList grammar in http://www.w3.org/TR/WebIDL/#idl-extended-attributes
 // Section 3.11
-[IntAttr=0, FloatAttr=3.14, StringAttr="abc", IdentifierAttr=_null, IdentifiersAttr=(_null, _const)]
+[IntAttr=0, FloatAttr=3.14, StringAttr="abc", IdentifierAttr=_null, IdentifiersAttr=(_null, _const), FloatList=(3.14)]
 interface IdInterface {};
 
 // Extracted from http://www.w3.org/TR/2016/REC-WebIDL-1-20161215/#Constructor on 2017-5-18 with whitespace differences
--- a/README.md
+++ b/README.md
@@ -711,6 +711,10 @@ Extended attributes look like this:
    * `"identifier-list"`
    * `"string"`
    * `"string-list"`
+   * `"decimal"`
+   * `"decimal-list"`
+   * `"integer"`
+   * `"integer-list"`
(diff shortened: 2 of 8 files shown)


test/syntax/idl/extended-attributes.webidl
```diff
@@ -7,7 +7,7 @@ interface ServiceWorkerGlobalScope : WorkerGlobalScope {
 
 // Conformance with ExtendedAttributeList grammar in http://www.w3.org/TR/WebIDL/#idl-extended-attributes
 // Section 3.11
-[IntAttr=0, FloatAttr=3.14, StringAttr="abc", IdentifierAttr=_null, IdentifiersAttr=(_null, _const)]
+[IntAttr=0, FloatAttr=3.14, StringAttr="abc", IdentifierAttr=_null, IdentifiersAttr=(_null, _const), FloatList=(3.14)]
```
TimothyGu: What would happen if we have a mixed attribute list? Like `[List=(8, 8.5)]`? I’d assume this should get parsed as a decimal-list but just wanted to contribute the test case.

saschanaz: Do we have a use case for mixed lists? The current implementation prevents mixed list as no such use case has been observed. For your specific example, I would suggest `[List=(8.0, 8.5)]`.

Sorry to respond too late 😥

TimothyGu: Your suggestion sounds fine – we'll report back if we happen to need mixed lists :wink:. Thanks for the clarification!

marcoscaceres:
Useful! Missed this being added to WebIDL! very cool.

saschanaz:
It's not really "added". The permissiveness has been there, while no specs really used it.

marcoscaceres:
Ah, explains why I thought I’d missed the discussion.

PR #457, opened 2020-02-03
test: run on Node.js 13

saschanaz:
Sad that we still don't have `lts/*` support. (https://github.com/actions/setup-node/issues/26)

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -9,7 +9,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [8.x, 10.x, 12.x]
+        node-version: [8.x, 10.x, 12.x, 13.x]
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/nodejs.yml
```diff
@@ -9,7 +9,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [8.x, 10.x, 12.x]
+        node-version: [8.x, 10.x, 12.x, 13.x]
```
marcoscaceres: should we ditch 8 and 10? I guess in a follow up.

saschanaz: We can, but could we do when we actually see any failures?

marcoscaceres: yeah, I'm ok with that.

PR #459, opened 2020-02-19
chore(package): Correctly add type field

ExE-Boss:
This adds the `type` field from <https://github.com/w3c/webidl2.js/pull/450> correctly while ensuring that <https://github.com/w3c/webidl2.js/pull/457> won’t break by renaming `webpack.config.js` to `webpack.config.cjs`.

> Per [the relevant Node.js documentation](https://nodejs.org/api/esm.html#esm_code_package_json_code_code_type_code_field). (See also [the blog post](https://medium.com/@nodejs/announcing-core-node-js-support-for-ecmascript-modules-c5d6dc29b663).)


saschanaz:
https://github.com/w3c/webidl2.js/pull/458 decided to revert and wait until it gets stable.

ExE-Boss:
I know, that’s why I opened this to do it correctly.

saschanaz:
You mean you will wait until it gets stable? 👀

ExE-Boss:
Well, the `type` field’s `"commonjs"` value is considered stable.

Same with the `.cjs` file extension, which has been supported for as long as **Node**’s support for extensionless files.

saschanaz:
I mean, [the whole ES module support is experimental](https://nodejs.org/api/esm.html#esm_ecmascript_modules), so I'll wait until it gets stable. `type: "commonjs"` would not be needed when we don't use native ES module support yet.

PR #505, opened 2020-11-18
chore(dependabot): update weekly

saschanaz:
This should reduce the distracting notifications 😄 (It's not that they are working actually, because of https://github.com/dependabot/dependabot-core/issues/2643 🙄)


saschanaz:
(I think this one is harmless)

PR #532, opened 2021-01-28
chore(README): Document the `default` field of arguments

### LINKED ISSUE #531: README does not reflect that arguments may have defaults
The argument type should consistently have a `default` field in the parsed output, and documentation should reflect that is is `null` (not "absent") when no default is specified.

cscott:
This patch closes #531 and includes:
- [ ] A relevant test


saschanaz (approved):
👍😍

PR #534, opened 2021-01-28
chore(README): Update documentation for the `special` field

### LINKED ISSUE #533: README documentation on the `special` field is out-of-date
One part of the README still uses the old separate `static`, `stringifier`, and `inherit` fields.  In other places it is stated that the `special` field is `null` in the default case, when in fact the `special` field is set to the empty string.

cscott:
One part of the README still uses the old separate static,
stringifier, and inherit fields. In other places it is stated that the
special field is null in the default case, when in fact the special
field is set to the empty string.

This patch closes #533 and includes:
- [ ] A relevant test

### DIFF
--- a/README.md
+++ b/README.md
@@ -536,7 +536,7 @@ An operation looks like this:
 The fields are as follows:
 
 * `type`: Always "operation".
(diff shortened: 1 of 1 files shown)


saschanaz (approved):
Great catch, thank you for doing this!

PR #536, opened 2021-01-28
chore(README): Update documentation for constant values

### LINKED ISSUE #535: README documentation for constant values incorrectly says sequences and booleans are strings
The README says, "The value of the given type, as a string." but in fact booleans and sequence types are not stored as strings.

cscott:
Constant values of type boolean and sequence types are not stored as strings.

This patch closes #535 and includes:
- [ ] A relevant test

### DIFF
--- a/README.md
+++ b/README.md
@@ -728,9 +728,9 @@ values, all of which have the following fields:
 
 * `type`: One of `"string"`, `"number"`, `"boolean"`, `"null"`, `"Infinity"`, `"NaN"`, `"sequence"` or `"dictionary"`.
 
-For `"string"`, `"number"`, `"boolean"`, and `"sequence"`:
+For `"boolean"`, `"string"`, `"number"`, and `"sequence"`:
 
-* `value`: The value of the given type, as a string. For sequence, the only possible value is `[]`.
+* `value`: The value of the given type.  For string and number types, the value is given as a string.  For booleans, the possible values are `true` and `false`. For sequence, the only possible value is `[]`.
 
 For `"Infinity"`:
 


README.md
```diff
@@ -728,9 +728,9 @@ values, all of which have the following fields:
 
 * `type`: One of `"string"`, `"number"`, `"boolean"`, `"null"`, `"Infinity"`, `"NaN"`, `"sequence"` or `"dictionary"`.
 
-For `"string"`, `"number"`, `"boolean"`, and `"sequence"`:
+For `"boolean"`, `"string"`, `"number"`, and `"sequence"`:
```
saschanaz, line 731: 👍😉

saschanaz (approved):

README.md
cscott: Yeah, sorry about the reordering, I didn't catch that when I cleaned the patch up.  I'd broken 'boolean' and 'sequence' out into separate bullet points at one point then decided that was overkill, but I guess I swapped the order a bit when I put them back.

Anyway, this is the last of the issues I found, for now at least.  https://github.com/cscott/WebIDL is running now (in PHP) and generates the same output as `webidl2.js`, modulo some details about parent fields and trivia.

saschanaz: Oops, I thought it's now alphabetically ordered but turns out it's not, but okay 😄

TIL the PIP library exists! I wonder one day we could build a wasm library and use it everywhere 👀

cscott: https://github.com/cscott/node-php-embed was my contribution to "use everywhere" ;)

Seriously, we @ wikimedia are working on improving the PHP DOM libraries, so expect to see more PHP/WebIDL work in the next few months as we drag PHP's ancient libxml bindings into the modern era.

Issue #545, opened 2021-02-17
Spurious validation error for dictionary with required member


foolip:
There appears to be a bug in `WebIDL2.validate` causing a spurious error for a dictionary which has a required member, where WebIDL2 omits errors as if the dictionary has no required members. To reproduce using webidl2.js 23.13.0:

```js
const WebIDL2 = require('webidl2');

const ast = WebIDL2.parse(`
  dictionary ThingInit : DoesNotExistInit {
    required DOMString foo;
  };
  
  [Exposed=Window]
  interface Thing {
    undefined foo(ThingInit init);
    undefined bar(ThingInit init);
  };
`);

const validations = WebIDL2.validate(ast);
console.log(validations);
```

This results in the error "Dictionary argument must be optional if it has no required fields". Odd things I found when trying to further minify the test case:
- If `ThingInit` does not inherit from `DoesNotExistInit`, the error goes away
- If there's only one of the `foo` and `bar` methods, the error goes away

This was found with [`background-fetch.idl` in webref](https://github.com/w3c/webref/blob/0e01ae81f42c84b2face603909a29e21cbd2dc5d/ed/idl/background-fetch.idl), where the problem is with the `BackgroundFetchEventInit` arguments.

@saschanaz FYI

foolip:
Note that it's not important for `DoesNotExistInit` to not exist, this dictionary definition also produces the error:
```
  dictionary BaseThingInit {
    required DOMString a;
  };

  dictionary ThingInit : BaseThingInit {
    required DOMString b;
  };
```

foolip:
I think I've spotted the problem (but not the solution) in `dictionaryIncludesRequiredField`...

foolip:
OK, it's a caching bug here:
https://github.com/w3c/webidl2.js/blob/2a356f263295787b6ad07aaa24952d48fcd79f8e/lib/validators/helpers.js#L65-L74

These two `return true` statements both leave `undefined` in the cache, and if the same dictionary is checked again, it will return `undefined` from the cache.

PR #546, opened 2021-02-17
Ensure cached values of dictionaryIncludesRequiredField are boolean

foolip:
The two `return true` statements would leave undefined in the cache, and
looking up the same dictionary again would then return undefined instead
of true. The bug manifests when the same dictionary appears as an
argument twice or more.

This patch closes https://github.com/w3c/webidl2.js/issues/545 and includes:
- [x] A relevant test
- [ ] A relevant documentation update &ndash; deemed not needed

### DIFF
--- a/lib/validators/helpers.js
+++ b/lib/validators/helpers.js
@@ -62,17 +62,19 @@ export function dictionaryIncludesRequiredField(dict, defs) {
   if (defs.cache.dictionaryIncludesRequiredField.has(dict)) {
     return defs.cache.dictionaryIncludesRequiredField.get(dict);
   }
-  defs.cache.dictionaryIncludesRequiredField.set(dict, undefined); // indeterminate
-  if (dict.inheritance) {
+  // Set cached result to indeterminate to short-circuit circular definitions.
+  // The final result will be updated to true or false.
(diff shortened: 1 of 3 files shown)


lib/validators/helpers.js
```diff
@@ -62,17 +62,19 @@ export function dictionaryIncludesRequiredField(dict, defs) {
...
+  if (!result && dict.inheritance) {
     const superdict = defs.unique.get(dict.inheritance);
     if (!superdict) {
-      return true;
-    }
-    if (dictionaryIncludesRequiredField(superdict, defs)) {
-      return true;
+      // Assume required members in the supertype if it is unknown.
+      result = true;
+    } else if (dictionaryIncludesRequiredField(superdict, defs)) {
```
saschanaz, line 74: I guess prettier will dislike this style, but okay for now.

saschanaz (approved):
Cool, thank you for doing this 👍👍

lib/validators/helpers.js
saschanaz: Ah no, it's not `return`, okay.

saschanaz:
v23.13.1 🚀

foolip:
Sweet, a new release :)

Issue #512, opened 2020-11-29
Consider adding a way to support browser-specific syntax


saschanaz:
Web IDL has received some changes that caused massive changes in web specs. [webidl-updater](https://github.com/saschanaz/webidl-updater) (via @autokagami) has been super helpful to complete the big migration, but so far there has been no tool for browser implementations. Supporting browser-specific syntax in ad-hoc way could boost them by letting them reuse the existing migration tool.

Work items would be:
* Expose factory methods (#490)
* Move writer.js into a `.write()` method in each class, so that it can call it for ad-hoc instances (#513)

cc @edgarchen for potential interest from Gecko side.

PR #556, opened 2021-04-07
refactor: expose as es module

saschanaz:
A maintenance for easier debugging when doing #512.

* This removes 10.x from CI workflow because of no module support, but since this is not dropping the UMD build, it's not a breaking change.
* Refactors `collect()` for module-compatible URL-based approach

Closes #459


saschanaz:
I think we can drop `dist/` from git and publish it only to npm, what do you think?

marcoscaceres (approved):
Nice and neat! ❤️

saschanaz:
It turns out that this is actually a breaking change, since the default Node.js behavior for importing CommonJS module in ESM context is to expose the module object as `default` item. I don't think we want to keep this behavior, so a wontfix and a major version bump.

package.json
```diff
@@ -28,11 +27,13 @@
...
     "acquire": "node test/util/acquire.js",
     "build": "npx webpack",
-    "build-debug": "npx webpack --mode none"
+    "build-debug": "npx webpack --mode none",
+    "prepublishOnly": "npm run build"
   },
   "repository": "git://github.com/w3c/webidl2.js",
   "main": "dist/webidl2.js",
-  "module": "index.js",
+  "exports": "./index.js",
```
ExE-Boss, line 35: Note that this makes it no longer possible to import anything other than `index.js`:
- <https://nodejs.org/api/packages.html#packages_package_entry_points>
- <https://nodejs.org/api/packages.html#packages_conditional_exports>
- <https://nodejs.org/api/packages.html#packages_exports>

saschanaz: Good point, and I think we never considered internal modules as stable, so still a good change I guess. Or were you using them?

ExE-Boss: It makes `dist/webidl2.js` unimportable in newer **Node** versions.

---

It should probably also be renamed to `dist/webidl2.cjs`, or add a `dist/package.json` file with:
```json
{ "type": "commonjs" }
```

saschanaz: Why should it be allowed to be imported in newer Node when there is `index.js`?

PR #558, opened 2021-04-07
feat(productions/attribute): validate

saschanaz:
This still needs to check generics inside unions but probably as a followup.

This is needed for Gecko since it still uses `attribute sequence<>`. The spec also doesn't block sequences in syntax level.

- [x] A relevant test


marcoscaceres (approved):
Lgtm

PR #562, opened 2021-04-08
chore: remove dist/ from git

saschanaz:
- [x] A relevant test: N/A
- [x] A relevant documentation update: N/A, the npm package will still have it


marcoscaceres (approved):

dontcallmedom:
one used to be able to load webidl2.js from `https://w3c.github.io/webidl2.js/dist/webidl2.js` - is there a new equivalent or is that no longer supported?

saschanaz:
> one used to be able to load webidl2.js from `https://w3c.github.io/webidl2.js/dist/webidl2.js` - is there a new equivalent or is that no longer supported?

Didn't know someone is using it 👀

One can still use `lib/index.js` but it's an ES module. Is it a viable option there?

saschanaz:
Or I wonder ReSpec can directly provide parsed IDLs for plugins. @sidvishnoi, do you have some thoughts?

sidvishnoi:
That could be supported via https://github.com/w3c/respec/pull/3320 (if). It would be more like exposing `utils.webidl = WebIDL2` though.

Meanwhile, can import `dist/` directly from npm instead of github (using unpkg or some other npm cdn):
https://unpkg.com/webidl2/dist/webidl2.js
https://unpkg.com/webidl2@24/dist/webidl2.js // semver

saschanaz:
On a second thought I think it's not very good idea since any breaking change from webidl2.js will make breaking change in ReSpec.

dontcallmedom:
> Meanwhile, can import `dist/` directly from npm instead of github

thanks, that's the perfect solution to my need; it might be useful to document it in the README

PR #565, opened 2021-04-15
fix(package): allow require() for commonjs use

saschanaz:
Okay, it turns out @exe-boss was right, and simple `exports` blocks CommonJS imports. I should have read https://github.com/w3c/webidl2.js/pull/556#discussion_r609879465 more carefully, especially https://nodejs.org/api/packages.html#packages_conditional_exports 😬

Somehow I thought Node would wrap ES modules to CommonJS as they do vise versa, but it doesn't.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules/
-dist/
+dist/*
+!dist/package.json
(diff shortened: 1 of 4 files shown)


marcoscaceres (approved):

marcoscaceres:
TIL also!

saschanaz:
Thanks for quick review ♥

PR #566, opened 2021-04-17
fix(tokeniser): update comment regex as the spec says

### LINKED ISSUE #559: Comment ends with **/ fails to parse
```webidl
/** what **/
```

This is a valid comment but the parser fails to parse it.

saschanaz:
The old regex was needlessly complex with an additional requirement (didn't allow `*` to exist without another preceding `*`).

This patch closes #559 and includes:
- [x] A relevant test
- [x] A relevant documentation update (N/A)


marcoscaceres (approved):
Nice little cleanup!

PR #568, opened 2021-04-18
refactor(writer): split to each module

### LINKED ISSUE #513: Split writer.js into a method on each class
It would be more modular in that way.

saschanaz:
This patch closes #513 and includes:
- [x] A relevant test (N/A)
- [x] A relevant documentation update (N/A, although #512 may need this to be documented.)


marcoscaceres (approved):
This is really awesome. I really like how clear everything is now.

PR #569, opened 2021-04-19
refactor: split consume(regex) as consumeType()

### LINKED ISSUE #313: Separate function to consume regex terminal symbols
`consume("decimal")` and `consume("float")` does fairly different thing but look too similar, being potentially vulnerable to bugs like #302. Maybe `regex("decimal")` would be good.

saschanaz:
This patch closes #313 and includes:
- [x] A relevant test (N/A)
- [x] A relevant documentation update (N/A, tokeniser has never been exposed as a stable API)

### DIFF
--- a/lib/productions/argument.js
+++ b/lib/productions/argument.js
@@ -36,7 +36,9 @@ export class Argument extends Base {
     if (!tokens.optional) {
       tokens.variadic = tokeniser.consume("...");
     }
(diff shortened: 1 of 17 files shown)


marcoscaceres (approved):
Lgtm

saschanaz:
Thanks as always 👍

PR #573, opened 2021-04-23
feat: support custom productions

### LINKED ISSUE #512: Consider adding a way to support browser-specific syntax

saschanaz:
This patch closes #512 and includes:
- [x] A relevant test

### DIFF
--- a/docs/custom-productions.md
+++ b/docs/custom-productions.md
@@ -23,15 +23,14 @@ declare function factory(
 The tokeniser object has the following form:
 
 ```ts
-enum TokenType {
-  "decimal",
-  "integer",
-  "identifier",
-  "string",
-  "whitespace",
-  "comment",
-  "other",
-}
+type TokenType =
+  | "decimal"
+  | "integer"
+  | "identifier"
+  | "string"
+  | "whitespace"
+  | "comment"
+  | "other";
 
 interface Tokeniser {
   /**
--- a/docs/custom-productions.md
+++ b/docs/custom-productions.md
@@ -59,7 +59,7 @@ interface Tokeniser {
    * Consumes a token if exists, or returns null.
    * Pass multiple token types to consume one of them.
    */
-  consumeType(...candidates: TokenType[]);
+  consumeKind(...candidates: TokenType[]);
   /**
    * Consumes an inline token if exists, or returns null.
    * Pass multiple token types to consume one of them.
--- a/docs/custom-productions.md
+++ b/docs/custom-productions.md
@@ -49,7 +49,7 @@ interface Tokeniser {
   /**
    * Checks whether a token exists without consuming it.
    */
-  probeType(type: TokenType): boolean;
+  probeKind(type: TokenType): boolean;
   /**
    * Checks whether an inline token exists without consuming it.
    * @example `probe("interface")`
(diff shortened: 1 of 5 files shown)


docs/custom-productions.md
```diff
@@ -0,0 +1,81 @@
+# Custom productions
+
```
marcoscaceres, line 2: This is still a bit vague (particularly the word "productions" ... don't need to change it, just needs an explanation). It might be good to expand this a little bit more to make it more clear when one might want to use this (i.e., maybe say it's good proprietary extensions).

docs/custom-productions.md
```diff
@@ -0,0 +1,81 @@
...
+   * Use this when you need to consume a custom token.
+   * @example `consumeIdentifier("legacycaller")`
+   */
+  consumeIdentifier(value: string);
+  /**
+   * Seeks to the position.
+   */
+  unconsume(position: number);
+}
+```
```
marcoscaceres, line 81: It might be good to have one concrete example of something simple that might use this.

lib/webidl2.js
```diff
@@ -58,6 +58,15 @@ function parseByTokens(tokeniser, options) {
   }
 
   function definition() {
+    if (options.productions) {
+      for (const production of options.productions) {
```
marcoscaceres, line 62: Q: maybe silly question, as I'm not 100% on how `productions` work, but seems odd that this would run through just one production, and then return after the first result is gotten? Is that correct?

marcoscaceres (changes requested):

docs/custom-productions.md
```diff
@@ -0,0 +1,81 @@
...
+```ts
+enum TokenType {
+  "decimal",
+  "integer",
+  "identifier",
+  "string",
+  "whitespace",
+  "comment",
+  "other",
+}
```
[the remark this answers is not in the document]
saschanaz: Ah you're right, I was confused with IDL enums. Thanks!

lib/webidl2.js
saschanaz: It must not be possible to match multiple productions at once, since it makes the grammar ambiguous. But here it does not actually check that and just assumes the first match is the unique match.

docs/custom-productions.md
saschanaz: I borrowed the word from [Web IDL spec](https://heycam.github.io/webidl/#idl-grammar) which is again from [the CS field](https://en.wikipedia.org/wiki/Production_(computer_science)). (Although I'm not 1000% sure I'm using the word here correctly 😅

Anyway I'll try making it clearer.

marcoscaceres: Ah, got it. I think it would be great to point to the Wikipedia article. That clarifies a lot!

marcoscaceres (approved):

PR #574, opened 2021-04-23
refactor: rename consumeType to consumeKind

saschanaz:
"type" has a special meaning in Web IDL (idl type), so probably better to use "kind" rather than overloading the word.

- [x] A relevant test

### DIFF
--- a/lib/productions/argument.js
+++ b/lib/productions/argument.js
@@ -37,7 +37,7 @@ export class Argument extends Base {
       tokens.variadic = tokeniser.consume("...");
     }
     tokens.name =
-      tokeniser.consumeType("identifier") ||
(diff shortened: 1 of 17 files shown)


marcoscaceres (approved):
"consume kind" make good sense.

saschanaz:
@marcoscaceres I'm now a little hesitant here, which would you prefer between consumeKind or consumeByKind? Somehow the former sounds like a property name 🤔

marcoscaceres:
I don't mind, tbh.. "consume" is a verb, so it still works well for the method name... but consumeByKind makes it even more clear. Either is fine by me.

saschanaz:
Cool, I'll merge this and see how people feels when we expose tokenisers.

PR #577, opened 2021-04-24
refactor: wrap EOF token

saschanaz:
This adds write() to eof object so that writer can stop special casing it. The tokeniser now also adds line/index fields to eof token for consistency with other tokens.

This removes `trivia` field from EOF. Theoretically it can be a breaking change, but trivia strings are not intended to be accessed outside of the writer function since #306. So I think it's safe to not mark it as breaking.

- [x] A relevant test (N/A)


marcoscaceres (approved):
LGTM!

Issue #570, opened 2021-04-19
Validate proper use of extended attributes


foolip:
Web IDL has a lot of requirements about which extended attributes may be used where. For example:
- interfaces: "[CrossOriginIsolated], [Exposed], [Global], [LegacyFactoryFunction], [LegacyNoInterfaceObject], [LegacyOverrideBuiltIns], [LegacyWindowAlias], and [SecureContext]."
- partial interfaces: "[CrossOriginIsolated], [Exposed], [LegacyOverrideBuiltIns], and [SecureContext]."
- interface mixins: "[CrossOriginIsolated], [Exposed], and [SecureContext]"
- namespace: "[CrossOriginIsolated], [Exposed], and [SecureContext]"
- dictionaries: no extended attributes

There's bound to be some violations of these rules in the wild, so it would be good to validate them.

(Required `[Exposed]` extended attributes are already validated.)

foolip:
I hacked something into webref's consistency tests to show the use for top-level definitions, which won't cover attributes and members. Here's the list of things that appear:
```
[Exposed] callback interface
[Exposed] dictionary
[Exposed] enum
[Exposed] interface
[Exposed] interface mixin
[Exposed] namespace
[Exposed] partial interface
[Exposed] partial namespace
[Global] interface
[HTMLConstructor] interface
[LegacyFactoryFunction] interface
[LegacyNamespace] interface
[LegacyNoInterfaceObject] interface
[LegacyOverrideBuiltIns] interface
[LegacyOverrideBuiltIns] partial interface
[LegacyTreatNonObjectAsNull] callback
[LegacyUnenumerableNamedProperties] interface
[LegacyWindowAlias] interface
[SecureContext] interface
[SecureContext] interface mixin
[SecureContext] partial interface
[Serializable] interface
[Transferable] interface
```

That confirms that there's some invalid use of at least `[Exposed]` for dictionaries and enums.

foolip:
I've sent https://github.com/WICG/portals/pull/265 and https://github.com/w3c/css-houdini-drafts/pull/1023 to fix the issues I spotted.

tidoust:
On top of validating the proper use of extended attributes, could there be an option to validate additional grammar constraints for extended attributes?

[...]

PR #578, opened 2021-04-24
BREAKING CHANGE: merge *-types as idl-type

### LINKED ISSUE #537: Promise<T> uses return-type in an inconsistent way; union types don't use type at all
*Usually* the rule seems to be that the top-most parent of a Type sets the `type` field all the way down.  `attribute-type`, for example, gets propagated down into sequence types, etc.

However, `union` types have the `type` property set to `null`, and these fields are not recursed into when setting the `attribute-type` value or whathaveyou.

Also `Promise<T>` sets the `type` property of the `T` type to `return-type`, and again that is robust and doesn't get overwritten with `attribute-type` --- *but only when T is `void`*.

In my PHP port, this leads to the somewhat baroque logic in the `setType` method here:
https://github.com/cscott/WebIDL/blob/main/src/Grammar.pegphp#L54
which has to take care not to overwrite this single type in the `Promise<void>` child set here:
https://github.com/cscott/WebIDL/blob/main/src/Grammar.pegphp#L764

In this repo's test cases, you can see the unusual behavior here, for `Promise<void>` (which gets `return-type` not `attribute-type`):
https://github.com/w3c/webidl2.js/blob/gh-pages/test/syntax/baseline/promise-void.json#L18
and here (`Promise<DOMString>` gets `attribute-type` not `return-type`):
https://github.com/w3c/webidl2.js/blob/gh-pages/test/syntax/baseline/generic.json#L60

And here is where you can see a union getting `null` as it's `type` instead of inheriting `attribute-type`:
https://github.com/w3c/webidl2.js/blob/gh-pages/test/syntax/baseline/uniontype.json#L18

This isn't necessarily a bug per se -- my PHP port can emulate this behavior consistently, and it's only two or three lines of code -- but it does seem... "non-orthogonal"... so I thought I'd mention it in case you want to clean it up.

I'd suggest consistently recursing into type children *including union types*, and getting rid of the `return-type` marker on `Promise<void>` and letting that be recursed into like all other types.

saschanaz:
This should make implementing #570 easier since every IDL type object will have the same type string.

This patch closes #537 and includes:
- [x] A relevant test

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,7 +223,7 @@ attached to a field called `idlType`:
 
 ```JS
 {
-  "type": "attribute-type",
+  "type": "idl-type",
   "generic": "",
   "idlType": "unsigned short",
   "nullable": false,
@@ -234,7 +234,7 @@ attached to a field called `idlType`:
 
 Where the fields are as follows:
 
-* `type`: String indicating where this type is used. Can be `null` if not applicable.
+* `type`: Always `"idl-type"`.
 * `generic`: String indicating the generic type (e.g. "Promise", "sequence").
 * `idlType`: String indicating the type name, or array of subtypes if the type is
   generic or a union.
@@ -266,7 +266,7 @@ Interfaces look like this:
 
 The fields are as follows:
 
-* `type`: Always "interface".
+* `type`: Always `"interface"`.
 * `name`: The name of the interface.
 * `partial`: `true` if the type is a partial interface.
 * `members`: An array of interface members (attributes, operations, etc.). Empty if there are none.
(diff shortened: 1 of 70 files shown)


marcoscaceres (approved):
Approving, but it almost makes `type: idl-type` kinda meaningless... should we maybe just remove it? What's the expected fallout from this breaking change on WPT and ReSpec (I think we use it in ReSpec too)?

saschanaz:
Closing, I guess something like `instanceof IDLType` should do the work.

PR #579, opened 2021-04-25
fix: change globalObject to `globalThis`

lucacasonato:
In ESM `this` is undefined at the top level. Instead to access the global object you must use `globalThis`.

This is causing some issues in Deno when running WPT idlharness tests, because all code is always executed in a module context.

### DIFF
--- a/webpack.config.cjs
+++ b/webpack.config.cjs
@@ -7,7 +7,7 @@ module.exports = {
     path: path.resolve(__dirname, "dist"),
     library: "WebIDL2",
     libraryTarget: "umd",
-    globalObject: "this"
+    globalObject: "globalThis"
   },
   mode: "production",
   devtool: "source-map",


saschanaz:
Hmm. Although this should probably be okay, could Deno use the ESM version of webidl2.js inside idlharness instead?

lucacasonato:
As you know the UMD version of webidl2.js is vendored into the wpt repo at https://github.com/web-platform-tests/wpt/blob/master/resources/webidl2/lib/webidl2.js. We could float a patch adjusting how idlharness tests are handled in our wpt fork (https://github.com/denoland/wpt), but this would require significant changes to `wpt serve` and how idlharness tests are structured. I don't think the overhead is worth it. If this patch doesn't get landed we will likely just float a patch adjusting the `this` to `globalThis` in the vendored webidl2.js.

lucacasonato:
For reference, the way we run the tests are to fetch the HTML page from `wpt serve`, then parse out the script imports, fetch them, and inline them all into a single file together with a custom test reporter. This is the third iteration of the WPT runner, and so far this has yielded the most accurate representation of what WPT test authors expect. Namely that all scripts run inside of the same context.

saschanaz:
My concern is that this is a syntactic breaking change, but I guess it should be okay as we expect this to run mainly on latest JS engines.

>this would require significant changes to wpt serve and how idlharness tests are structured. I don't think the overhead is worth it.

I hope WPT ultimately migrate to ESM, so I'd say it's kinda worth it 😄

saschanaz (approved):

saschanaz:
Since there already is a pending breaking change, let's merge this and see how it goes.

PR #582, opened 2021-04-26
Document return types of parse() and validate()

foolip:


foolip:
I don't really know what I'm doing here. I think this is correct, but I don't know how things would fail if it were wrong.

lib/validator.js
```diff
@@ -86,7 +86,8 @@ function flatten(array) {
 }
 
 /**
- * @param {*} ast AST or array of ASTs
+ * @param {*[]} ast AST or array of ASTs
+ * @return {error[]} validation errors
```
saschanaz, line 90: `*[]` is better than `*` but `error[]` is not since there is no type named `error`. Maybe also `*[]` for now?

saschanaz (changes requested):

saschanaz (approved):
Adjusted a bit myself, merging.

PR #584, opened 2021-04-27
Implement a merge() method to apply partials and mixins

### LINKED ISSUE #581: New method to merge partials and mixins
There's code using webidl2.js to parse IDL snippets and then applying partials/mixins in at least 3 different places:
- [idlharness.js](https://github.com/web-platform-tests/wpt/blob/e6f0102f15039f11d79180b0bded62cf0633ce73/resources/idlharness.js#L700-L702)
- [webref](https://github.com/w3c/webref/blob/b3c9b370c6a17c2748bd0f55ce32a18820305479/test/idl/consistency.js#L313-L315)
- [mdn-bcd-collector](https://github.com/foolip/mdn-bcd-collector/blob/c002a7f18bf2712fdf6c5ae6c4b919db280eaf64/build.js#L189)

None of them are entirely correct in the handling of `[Exposed]` and updating them individually is a nuisance.

How about a `compile(ast)`, `merge(ast)`, or `simplify(ast)` method that folds in partials and mixins as part of webidl2.js?

foolip:
This patch closes #581 and includes:
- [ ] A relevant test

### DIFF
--- a/lib/merge.js
+++ b/lib/merge.js
@@ -1,0 +1,140 @@
+import { ExtendedAttributes } from "./productions/extended-attributes.js";
+import { Tokeniser } from "./tokeniser.js";
+
+// Remove this once all of our support targets expose `.flat()` by default
+function flatten(array) {
+  if (array.flat) {
+    return array.flat();
+  }
+  return [].concat(...array);
+}
+
+// https://heycam.github.io/webidl/#own-exposure-set
+function getOwnExposureSet(node) {
+  const exposedAttr = node.extAttrs.find((a) => a.name === "Exposed");
+  if (!exposedAttr) {
+    return null;
+  }
+  const exposure = new Set();
+  const { type, value } = exposedAttr.rhs;
+  if (type === "identifier") {
+    exposure.add(value);
+  } else if (type === "identifier-list") {
+    for (const ident of value) {
+      exposure.add(ident.value);
+    }
+  }
+  return exposure;
+}
+
+/**
+ * @param {Set?} a a Set or null
+ * @param {Set?} b a Set or null
+ * @return {Set?} a new intersected set, one of the original sets, or null
+ */
+function intersectNullable(a, b) {
+  if (a && b) {
+    const intersection = new Set();
+    for (const v of a.values()) {
+      if (b.has(v)) {
+        intersection.add(v);
+      }
+    }
+    return intersection;
+  }
+  return a || b;
+}
+
+/**
+ * @param {Set?} a a Set or null
+ * @param {Set?} b a Set or null
+ * @return true if a and b have the same values, or both are null
+ */
+function equalsNullable(a, b) {
+  if (a && b) {
+    if (a.size !== b.size) {
+      return false;
+    }
+    for (const v of a.values()) {
+      if (!b.has(v)) {
+        return false;
+      }
+    }
+  }
+  return a === b;
+}
+
+/**
+ * @param {Container} target definition to copy members to
+ * @param {Container} source definition to copy members from
+ */
+function copyMembers(target, source) {
+  const targetExposure = getOwnExposureSet(target);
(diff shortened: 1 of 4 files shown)


foolip:
This is pretty far from done, but sending it as a slightly-working checkpoint, in order to comment on a few issues.

lib/merge.js
```diff
@@ -0,0 +1,140 @@
...
+    } else if (dfn.type === "includes") {
+      includes.push(dfn);
+    } else if (dfn.name) {
+      dfns.set(dfn.name, dfn);
+    } else {
+      throw new Error(`definition with no name`);
+    }
+  }
+
+  // merge partials (including partial mixins)
```
foolip, line 123: The idea here is to first merge partials and then merge mixins, performing basically the same steps each time. But this will have the downside of copying members and merging extended attributes twice in the case of a partial interface mixin. (The mixin step isn't here yet.)

Performance isn't much of a concern, but I wonder if this is overgeneralized, in particular it becomes hard to judge whether https://heycam.github.io/webidl/#dfn-exposure-set is followed.

The alternative would be to first collect all includes statements and then to copy members directly from their source to their final destination.

lib/merge.js
```diff
@@ -0,0 +1,140 @@
...
+      Object.getPrototypeOf(orig),
+      Object.getOwnPropertyDescriptors(orig)
+    );
+
+    if (!equalsNullable(targetExposure, copyExposure)) {
+      let value = Array.from(copyExposure.values()).join(",");
+      if (copyExposure.size !== 1) {
+        value = `(${value})`;
+      }
+      copy.extAttrs = ExtendedAttributes.parse(
```
foolip, line 93: @saschanaz I found it quite hard to get this right, and I didn't. The code here ultimately needs to copy any extended attributes from partial/mixin to its members, and to merge `[Exposed]` as part of that. That might end up adding, removing or modifying the original list of extended attributes. It seems like the only way to do this will be to carefully modify the original tokens, serialize it, and then `ExtendedAttributes.parse` the string. Are the helpers that will make this easier, or should I go about it differently?

lib/merge.js
```diff
@@ -0,0 +1,140 @@
...
+    target.members.push(copy);
+  }
+}
+
+/**
+ * @param {*[]} ast AST or array of ASTs
+ * @return {*[]}
+ */
+export function merge(ast) {
+  const dfns = new Map();
```
saschanaz, line 107: Probably good to reuse this function:

https://github.com/w3c/webidl2.js/blob/49c514f0bc1aba6919381ed11fbdddcfc05efc48/lib/validator.js#L24

lib/merge.js
saschanaz: >in particular it becomes hard to judge whether https://heycam.github.io/webidl/#dfn-exposure-set is followed.

Not sure I follow this part, in what situation would it be problematic?

test/merge.js
```diff
@@ -0,0 +1,72 @@
+import expect from "expect";
+import { parse, write, merge } from "webidl2";
+
+// Collapse sequences of whitespace to a single space.
+function collapse(s) {
+  return s.trim().replace(/\s+/g, " ");
+}
+
+expect.extend({
+  toMergeAs(received, expected) {
```
saschanaz, line 10: Wow, didn't know `expect` supports extension this way 👍

foolip: Me neither, I found out after I had repeated myself a lot and wanted a helper function :)

lib/merge.js
foolip: I mean if the implementation isn't similar to the algorithm in the spec, it's harder to know if it get the same result in all cases. I think the two-step approach would be equivalent, but I'm not certain. The rules about whether the extended attributes of the member have to be a subset of the container's or if the intersection is used differ at least, so a unified implementation would allow for more invalid stuff.

PR #620, opened 2021-09-06
Fix online checker link

CYBAI:
Currently, the `checker` link refers to the [`checker` folder](https://github.com/w3c/webidl2.js/tree/main/checker) in this repository but I wonder maybe the https://w3c.github.io/webidl2.js/checker/ online checker might be the right one.

- [ ] A relevant test


saschanaz (approved):
I'd prefer to keep being compatible with forks but seemingly there is no such way. 😬

PR #634, opened 2021-11-02
feat: support `[Exposed=*]`

saschanaz:
Corresponds to https://github.com/whatwg/webidl/pull/526

- [x] A relevant test


test/syntax/baseline/exposed-asterisk.json
```diff
@@ -0,0 +1,24 @@
+[
+    {
+        "type": "interface",
+        "name": "X",
+        "inheritance": null,
```
saschanaz, line 5: Hmm? Is `null` deprecated in JSON?

marcoscaceres: Don’t think so? Weird.

marcoscaceres (approved):

test/syntax/baseline/exposed-asterisk.json
ExE-Boss: > Is `null` deprecated in JSON?

Almost certainly not, because as far as I understand it, **JSON** syntax is forever set in stone, which unfortunately means that comments will almost certainly never be supported, even though they existed in the initial drafts.

saschanaz: Yes, that's my understanding. Not sure why the red highlight 🤔. Also not sure where to report.

ExE-Boss: Looks like it just got fixed in <https://github.com/Nixinova/NovaGrammars/pull/7>, which is the project @GitHub uses for highlighting JSON syntax: <https://github.com/github/linguist/tree/master/vendor>.

PR #636, opened 2021-11-06
fix(checker): use `.value` instead of `.textContent` for textareas

saschanaz:
`.textContent` is ignored when the value is modified.

### DIFF
--- a/checker/index.html
+++ b/checker/index.html
@@ -1,91 +1,85 @@
 <!DOCTYPE html>
-<html>
+<meta charset="utf-8">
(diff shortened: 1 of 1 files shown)


marcoscaceres (approved):
Any reason to change the spaces?

saschanaz:
No, it's just done by prettier

saschanaz:
Ah, actually https://github.com/w3c/webidl2.js/pull/636/commits/5c8c43d989802ec56bd08937237788909e7077d3

PR #673, opened 2022-05-19
Allow same named static and nonstatic operations

### LINKED ISSUE #672: "no-cross-overload" should not fire on static and regular operations with same identifier
Web IDL was recently updated to allow static operations to have the same name as regular operations, see https://github.com/whatwg/webidl/issues/1097

The checker seems happy with that as long as these operations are defined in the same base definition. However, the checker reports a `no-cross-overload` error when the definitions are scattered across partials or mixins. For instance, the checker reports the error for the following IDL:

```webidl
[Exposed=*]
interface Test {
  undefined operation();
};

partial interface Test {
  static undefined operation();
};
```

Since operations are not of the same kind (regular vs. static), they should not be considered as [overloaded](https://webidl.spec.whatwg.org/#idl-overloading) operations, and I think the preceding IDL is valid.

Real world example is the recent update to fetch to add a static `Response.json` operation: https://github.com/whatwg/fetch/commit/b3bfd0c877fd09da8c315b43cf1dd858f3ac1612. In that case, the static operation is on the base mixin interface `Body` and the regular operation on `Response`. The Web IDL checker complains with the same "no-cross-overload" error.

saschanaz:
This patch closes #672 and includes:
- [x] A relevant test
- [x] A relevant documentation update (N/A)

### DIFF
--- a/lib/validators/interface.js
+++ b/lib/validators/interface.js
@@ -1,22 +1,35 @@
 import { validationError } from "../error.js";
 
+/**
+ * @param {import("../validator.js").Definitions} defs
+ * @param {import("../productions/container.js").Container} i
+ */
 export function* checkInterfaceMemberDuplication(defs, i) {
-  const opNames = new Set(getOperations(i).map((op) => op.name));
+  const opNames = groupOperationNames(i);
   const partials = defs.partials.get(i.name) || [];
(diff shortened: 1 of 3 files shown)


marcoscaceres (approved):
Looks great 👍

PR #731, opened 2023-04-11
feat: allow adding additional members to production parse methods

pyoor:
This PR adds the `extMembers` key to the production parse method's option bag.  The `extMembers` key is used to add additional members that may exist during parsing.  This is especially useful for allowing custom productions expand existing productions.

Additionally, this PR adds `extensions` to the parse method so that users can add additional allowedMembers on a per production basis.


lib/webidl2.js
```diff
@@ -32,28 +44,42 @@ function parseByTokens(tokeniser, options) {
     const callback = consume("callback");
     if (!callback) return;
     if (tokeniser.probe("interface")) {
-      return CallbackInterface.parse(tokeniser, callback);
+      return CallbackInterface.parse(tokeniser, callback, {
+        ...options?.extensions?.["callback-interface"],
```
saschanaz, line 48: callbackInterface maybe?

lib/webidl2.js
```diff
@@ -72,11 +98,11 @@ function parseByTokens(tokeniser, options) {
       callback() ||
       interface_() ||
       partial() ||
-      Dictionary.parse(tokeniser) ||
+      Dictionary.parse(tokeniser, { ...options?.extensions?.dictionary }) ||
```
saschanaz, line 101:
```suggestion
      Dictionary.parse(tokeniser, options?.extensions?.dictionary) ||
```

Probably no need to spread at all. Same for namespace below.

test/extensions.js
```diff
@@ -0,0 +1,20 @@
...
+import expect from "expect";
+import { Attribute } from "../lib/productions/attribute.js";
+
+const extensions = {
+  "callback-interface": {
+    extMembers: [[Attribute.parse]],
+  },
+  namespace: {
+    extMembers: [[Attribute.parse, { readonly: false }]],
+  },
```
saschanaz, line 11: I think it's better to have unit tests, like https://github.com/w3c/webidl2.js/blob/main/test/custom-production.js.

saschanaz (changes requested):

test/extensions.js
```diff
@@ -0,0 +1,50 @@
+"use strict";
```
saschanaz, line 1: We don't need `"use strict"` here, it's the default in ES modules!

saschanaz: Oh then I found the test I gave you still has it. Anyway it's not needed 😛

pyoor: Removed in commit [f12005e](https://github.com/w3c/webidl2.js/pull/731/commits/f12005ee66c058dd155c628d63168043b04de1b1)

test/custom-production.js
```diff
@@ -1,7 +1,40 @@
...
 import expect from "expect";
+
 import { parse, write } from "webidl2";
+import { Base } from "../lib/productions/base.js";
+import {
+  autoParenter,
+  type_with_extended_attributes,
+} from "../lib/productions/helpers.js";
+
+class CustomAttribute extends Base {
```
saschanaz, line 10: Good, can you replace the existing test below to use this? I think this one is better than `const customProduction` in it.

test/util/collect.js
```diff
@@ -7,7 +7,7 @@ import { diff } from "jsondiffpatch";
  * Collects test items from the specified directory
  * @param {string} base
  */
-export function* collect(base, { expectError, raw } = {}) {
+export function* collect(base, { expectError, raw, extensions = {} } = {}) {
```
saschanaz, line 10: (I guess this is not needed anymore)

pyoor: Addressed in [1b29589](https://github.com/w3c/webidl2.js/pull/731/commits/1b2958989a9fd81796d82b668b112222ef0d1648).

test/custom-production.js
pyoor: Addressed in [9cfc34c](https://github.com/w3c/webidl2.js/pull/731/commits/9cfc34c560c39960f13eb06b16c183e01bfd2165).

saschanaz:
Oh, oops. Could you open this again targeting `main` branch? 👀

PR #732, opened 2023-04-13
feat: expose non-top-level production classes

saschanaz:
This does not expose top level productions for now, since the parse function of them have some weird internal-only behavior, e.g. `Interface.parse` does not parse `interface` keyword by itself. Technically same for `Attribute` and `Operation`, but they are usable at least.

This patch includes:
- [x] A relevant test (N/A, just exposes things)


saschanaz:
Turns out I pushed the wrong button.

marcoscaceres (approved):

PR #745, opened 2023-06-14
 feat: allow adding additional members to production parse methods

pyoor:
This PR adds the extMembers key to the production parse method's option bag. The extMembers key is used to add additional members that may exist during parsing. This is especially useful for allowing custom productions expand existing productions.

Additionally, this PR adds extensions to the parse method so that users can add additional allowedMembers on a per production basis.


lib/webidl2.js
```diff
@@ -12,11 +12,22 @@ import { CallbackInterface } from "./productions/callback-interface.js";
 import { autoParenter } from "./productions/helpers.js";
 import { Eof } from "./productions/token.js";
 
+/** @typedef {'callbackInterface'|'dictionary'|'interface'|'mixin'|'namespace'} ExtendableInterfaces */
+/** @typedef {Object.<'extMembers', import("./productions/container.js").AllowedMember[] >} Extension */
+/** @typedef {Object.<[ExtendableInterfaces], Extension>} Extensions */
```
saschanaz, line 17: This is currently not recognized as valid type expressions.

```suggestion
/** @typedef {{ extMembers?: import("./productions/container.js").AllowedMember[]}} Extension */
/** @typedef {Partial<Record<ExtendableInterfaces, Extension>>} Extensions */
```

I know it's weird to mix TypeScript into JSDoc, but it's what we have. 🥲

lib/productions/container.js
```diff
@@ -16,6 +16,19 @@ function inheritance(tokeniser) {
...
+ * @param {import("../tokeniser.js").Tokeniser} tokeniser
+ * @param {...*} args
+ */
+
+/**
+ * A parser callback and optional option object.
+ * @typedef AllowedMember
+ * @type {[parserCallback, object?]}
+ */
+
```
saschanaz, line 31:
```suggestion
/**
 * Parser callback.
 * @callback ParserCallback
 * @param {import("../tokeniser.js").Tokeniser} tokeniser
 * @param {...*} args
 */

/**
 * A parser callback and optional option object.
 * @typedef AllowedMember
 * @type {[ParserCallback, object?]}
 */

```

PascalCase for types.

saschanaz (changes requested):
Some nits:

lib/productions/callback-interface.js
```diff
@@ -16,8 +20,9 @@ export class CallbackInterface extends Container {
       tokeniser,
       new CallbackInterface({ source: tokeniser.source, tokens }),
       {
-        inheritable: !partial,
+        inheritable: !!inheritable,
```
saschanaz, line 23: Can we make this change separately? We should either change all `partial` to `inheritable` everywhere or keep it everywhere, and making that change here sounds like out-of scope.

pyoor: Wasn't the issue here that callback interface doesn't support `partial` at all?

saschanaz: Oh hmm, I think we talked about it somewhere but can't find it anywhere, do you remember? I now wonder why we need it at all, neither the spec nor the mozilla-central uses inheritance for callback interfaces, right?

saschanaz: Ah I found it, and I think we should just remove `inheritable`s here.

PR #747, opened 2023-06-22
feat: parse SharedArrayBuffer and validate AllowShared

### LINKED ISSUE #746: Give SharedArrayBuffer a dedicated type
See https://github.com/whatwg/webidl/pull/1311.

saschanaz:
This patch closes #746 and includes:
- [x] A relevant test

### DIFF
--- a/lib/productions/type.js
+++ b/lib/productions/type.js
@@ -199,16 +199,17 @@ export class Type extends Base {
       // https://webidl.spec.whatwg.org/#idl-annotated-types
       for (const extAttrs of [this.extAttrs, this.parent?.extAttrs]) {
         for (const extAttr of extAttrs) {
-          if (extAttr.name === "AllowShared") {
-            const message = `\`[AllowShared] BufferSource\` is now replaced with AllowSharedBufferSource.`;
-            yield validationError(
-              this.tokens.base,
-              this,
-              "migrate-allowshared",
-              message,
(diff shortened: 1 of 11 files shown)


lib/productions/type.js
```diff
@@ -194,6 +194,25 @@ export class Type extends Base {
   *validate(defs) {
     yield* this.extAttrs.validate(defs);
 
+    if (this.idlType === "BufferSource") {
+      // XXX: For now this is a hack. Consider moving parents' extAttrs into types as the spec says:
+      // https://webidl.spec.whatwg.org/#idl-annotated-types
+      for (const extAttrs of [this.extAttrs, this.parent?.extAttrs]) {
```
marcoscaceres, line 200: Could this be `[this.extAttrs, this.parent?.extAttrs].flat()` to avoid the second for loop?

marcoscaceres: yeah, maybe also... to flatten to code a little more:
``` 
for (const extAttr of [this.extAttrs, this.parent?.extAttrs]) {
   if (extAttr.name !== "AllowShared") continue;
   // the rest... 
}
```

marcoscaceres (approved):
Just a style nit.

lib/productions/type.js
saschanaz: That way the autofix won't know which array includes the attr and it'll need to rescan 😬

PR #771, opened 2024-02-13
Update package.json with license: MIT

### LINKED ISSUE #702: Switch license
Could we switch license to W3C Software and Doc license? It would make it easier for me to contribute again.

saschanaz:
This field doesn't mean much when LICENSE is still in MIT. (i.e. this library has been under MIT license forever regardless of this field.)

This patch closes #702

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
   "engines": {
     "node": ">= 14"
   },
-  "license": "W3C",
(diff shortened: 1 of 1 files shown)


saschanaz:
Oh oops, but the issue says go to the other way.

PR #775, opened 2024-03-26
feat: add support for parsing `async_sequence<T>` type

lucacasonato:
Adds support for `async iterable<T>` (https://github.com/whatwg/webidl/pull/1397).


lib/productions/attribute.js
```diff
@@ -75,7 +75,9 @@ export class Attribute extends Base {
     yield* this.extAttrs.validate(defs);
     yield* this.idlType.validate(defs);
 
-    if (["sequence", "record"].includes(this.idlType.generic)) {
+    if (
+      ["async iterable", "sequence", "record"].includes(this.idlType.generic)
```
saschanaz, line 79: It indeed doesn't make sense but the spec PR does not fix https://whatpr.org/webidl/1397/402a886...0402e85.html#idl-attributes "The type of the attribute ..." part.

lucacasonato: The spec PR disallows it outright in any "return to JS position" because it does not define IDL->JS conversion (only JS->IDL).

saschanaz: I'd say yes as sequence being attribute is disallowed multiple times in the spec (at the attribute definition and also at the sequence definition).

lucacasonato:
This is now ready for review.

saschanaz (changes requested):
Mostly fine, but I do not see the change corresponding to constants:

> Async iterables must not be used as the type of an attribute or constant.

The Constant interface currently has no validate() but it'll be called automatically by Container when you add it.

saschanaz:
> Mostly fine, but I do not see the change corresponding to constants:
> 
> > Async iterables must not be used as the type of an attribute or constant.
> 
> The Constant interface currently has no validate() but it'll be called automatically by Container when you add it.

I should probably do this myself...

lib/productions/attribute.js
lukewarlow, line 79: Per https://github.com/whatwg/webidl/pull/1489 this PR should be updated.

saschanaz (approved):

saschanaz:
Thanks!

PR #777, opened 2024-04-03
feat: add Float16Array

bakkot:
Adds support for Float16Array from the [Float16Array proposal for JS](https://github.com/tc39/proposal-float16array); corresponding webidl change at https://github.com/whatwg/webidl/pull/1398.


saschanaz (approved):
LGTM, I'll merge when the upstream PR merges.

